PR #31, opened 2021-09-24
Add character counter for input and textarea

tajakobsen:
Relates to: MT-327

### DIFF
--- a/src/ts/types.d.ts
+++ b/src/ts/types.d.ts
@@ -2,3 +2,11 @@ export interface UIError {
   fieldName: string;
   message: string;
 }
+
+export type Action<Node extends HTMLElement> = (
+  node: Node,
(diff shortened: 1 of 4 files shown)


src/ts/types.d.ts
```diff
@@ -2,3 +2,11 @@ export interface UIError {
   fieldName: string;
   message: string;
 }
+
+export type Action<Node extends HTMLElement> = (
```
lha047, line 6: Doesnt Svelte provide a type for this? 🤔

lha047 (approved):
Nice work :D

src/ts/types.d.ts
tajakobsen: Couldn't find any type for it

PR #35, opened 2021-09-28
Mt 331 transport cards update

lha047:
Relates to MT-331


stories/visualStyle/Typography.stories.svelte
```diff
@@ -32,10 +32,44 @@
     <p>{paragraph}</p>
     <p class="text-small">{small}</p>
     <form action="">
-      <label for="input">{label}</label>
+      <label for="input" class="form-label">{label}</label>
       <input id="input" class="form-field" type="text" />
     </form>
     <a href="">{link}</a>
+    <a
```
tajakobsen, line 39: I'm a bit unsure if this belongs under typography

src/scss/global/_links.scss
```diff
@@ -41,19 +43,22 @@ a {
     }
   }
   &.transport-card {
-    display: flex;
-    flex-direction: column;
+    display: grid;
```
tajakobsen, line 46: What was the gain in switching to `display: grid`?

tajakobsen (changes requested):
Good job! Just a few small feedback points.

stories/visualStyle/Typography.stories.svelte
lha047: I wanted to show the font size of these components. To have a comparison with the other used font sizes

tajakobsen (approved):
Looks good. Merging to main.

PR #42, opened 2021-10-05
Mt 124 legal guidance

itemodahl:


src/svelte/Disclosure.svelte
```diff
@@ -55,9 +56,9 @@
 
 <div class="disclosure">
   {#if onServer}
-    <h3 class="disclosure-header">{title}</h3>
+    <h3 class="disclosure-header {size}">{title}</h3>
```
lha047, line 59: Im wondring if we should allow for the tag to be configuable and not just the styling 🤔

lha047 (changes requested):
Good job :D 
Im not to sure about whats right on the disclosure heading. Can we have a discussion on it?

src/scss/components/_disclosure.scss
```diff
@@ -35,6 +35,12 @@
         transform: rotate(-180deg);
       }
     }
+
+    &.large {
```
lha047, line 39: We dont need this now that we have propert tags for styling

src/svelte/Disclosure.svelte
```diff
@@ -10,6 +10,8 @@
 
   export let loadJs = true;
   export let title: string;
+  export let size: string;
```
lha047, line 13: Do we need this now that we have proper tags to handle headings?

lha047 (changes requested):
Nice :D Just one more thing...

lha047 (approved):
Looks good :D

PR #51, opened 2021-10-20
Mt 322 jump to content link

itemodahl:


stories/SkipLink.stories.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+<Meta
+  title="Components/Skip Link"
+  args={{
+    label: 'Hopp til hovedinnholdet',
+  }}
+/>
+
+<Story name="Hidden focusable link" let:label>
+  <a href="" class="visually-hidden-focusable">{label}</a>
```
tajakobsen, line 13: You actually get the wrong semantics if `href` doesn't have a value. It's a bit weird.

```suggestion
  <a href="#" class="visually-hidden-focusable">{label}</a>
```

src/scss/utilities/_layout.scss
```diff
@@ -75,6 +75,29 @@
   clip-path: inset(50%);
 }
 
+.visually-hidden-focusable {
+  background: var(--color-primary);
+  color: white;
+  width: 100%;
+  height: auto;
+  position: fixed;
```
tajakobsen, line 83: We get some weird margin on the left that is not on the right. I would consider adding:

```css
  left: 0;
```

tajakobsen (approved):
Looks good. Closing and merging to main

PR #67, opened 2021-11-11
MT-452: Collaps breadcrumbs if it is longer than 3 elements

itemodahl:
MT-452

### DIFF
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -112,6 +112,21 @@ $button-colors: 'error' !default;
     }
   }
 
+  &--link {
+    border: 0 none;
+    min-width: 44px;
+    background: transparent;
+    color: var(--color);
+    text-decoration: underline;
+    min-height: 17px;
+    margin: 0px;
+    padding: 0;
+    &:hover {
+      background: transparent;
+      color: var(--color);
+    }
+  }
+
   &__icon {
     width: 1em;
     height: 1em;
@@ -155,6 +170,14 @@ $button-colors: 'error' !default;
     }
     100% {
       transform: rotate(360deg);
+    }
+  }
+
+  &.forward-arrow {
+    &::after {
+      content: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iNiIgaGVpZ2h0PSIxMCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48cGF0aCBkPSJNLjIgMS40MyAzLjY2MSA1IC4yIDguNTdsMS4wNjYgMS4wOTdMNS44IDUgMS4yNjYuMzMzLjIgMS40M1oiIGZpbGw9IiM0NjQ1NDUiLz48L3N2Zz4=');
+      display: inline-block;
+      margin-left: var(--spacer-x-small);
     }
   }
 }
--- a/src/scss/components/_buttons.scss
+++ b/src/scss/components/_buttons.scss
@@ -114,12 +114,12 @@ $button-colors: 'error' !default;
 
   &--link {
     border: 0 none;
-    min-width: 44px;
     background: transparent;
     color: var(--color);
     text-decoration: underline;
-    min-height: 17px;
-    margin: 0px;
+    min-width: auto;
+    min-height: auto;
+    margin: 0;
     padding: 0;
     &:hover {
       background: transparent;
@@ -174,10 +174,10 @@ $button-colors: 'error' !default;
   }
 
   &.forward-arrow {
-    &::after {
+    &::before {
       content: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iNiIgaGVpZ2h0PSIxMCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48cGF0aCBkPSJNLjIgMS40MyAzLjY2MSA1IC4yIDguNTdsMS4wNjYgMS4wOTdMNS44IDUgMS4yNjYuMzMzLjIgMS40M1oiIGZpbGw9IiM0NjQ1NDUiLz48L3N2Zz4=');
       display: inline-block;
-      margin-left: var(--spacer-x-small);
+      margin-right: var(--spacer-x-small);
     }
   }
 }
--- a/src/svelte/Breadcrumbs.svelte
+++ b/src/svelte/Breadcrumbs.svelte
@@ -1,18 +1,58 @@
 <script lang="ts">
+  import {onMount} from "svelte";
+
   interface Breadcrumbs {
-    items: Array<{url: string; title: string}>;
-    ariaLabel?: string;
+    items: Array<{url: string; title: string}>
+    ariaLabel?: string
+    showAllAriaLabel?: string
   }
 
   export let breadcrumbs: Breadcrumbs = {items: []};
+  export let loadJs = false
+
   let ariaLabel = breadcrumbs.ariaLabel ?? 'breadcrumbs';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'show all'
+  let spread = true
+  $: breadcrumbData = breadcrumbs.items
+
+  function spreadItems(){
+    spread = true
+    breadcrumbData = breadcrumbs.items
+  }
+
+  if(loadJs && breadcrumbs.items.length > 3) {
+    onMount(() => {
+      spread = false
+      breadcrumbData = [
+        breadcrumbs.items[0],
+        breadcrumbs.items[1], {
+          title: '...',
+          text: '...',
+          active: false,
+          url: '#',
+          type: 'shorthand',
+          collapsed: true
+        },
+        breadcrumbs.items[breadcrumbs.items.length -1]]
+    })
+  }
 </script>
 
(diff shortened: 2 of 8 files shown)


src/scss/components/_buttons.scss
```diff
@@ -112,6 +112,21 @@ $button-colors: 'error' !default;
     }
   }
 
+  &--link {
+    border: 0 none;
+    min-width: 44px;
```
tajakobsen, line 117: I don't think this is necessary. WCAG 2.5.5 is AAA, and I'm unsure if it applies to breadcrumbs, so I would probably not add this.

I would ensure that the elipsis button is large enough though.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,18 +1,58 @@
...
+    showAllAriaLabel?: string
   }
 
   export let breadcrumbs: Breadcrumbs = {items: []};
+  export let loadJs = false
+
   let ariaLabel = breadcrumbs.ariaLabel ?? 'breadcrumbs';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'show all'
+  let spread = true
+  $: breadcrumbData = breadcrumbs.items
```
tajakobsen, line 16: `breadcrumbData` does not need to be responsive
```suggestion
  let breadcrumbData = breadcrumbs.items
```

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,18 +1,58 @@
...
+    items: Array<{url: string; title: string}>
+    ariaLabel?: string
+    showAllAriaLabel?: string
   }
 
   export let breadcrumbs: Breadcrumbs = {items: []};
+  export let loadJs = false
+
   let ariaLabel = breadcrumbs.ariaLabel ?? 'breadcrumbs';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'show all'
```
tajakobsen, line 14: I would probably use norwegian words as fallback here, since this will most likely be used on norwegian pages.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,18 +1,58 @@
...
+      spread = false
+      breadcrumbData = [
+        breadcrumbs.items[0],
+        breadcrumbs.items[1], {
+          title: '...',
+          text: '...',
+          active: false,
+          url: '#',
+          type: 'shorthand',
+          collapsed: true
```
tajakobsen, line 34: The values in this object is almost never used. Did you consider just using a constant string instead. It would make it easier to read.

```typescript
const BUTTON_ELIPSIS = Symbol();

breadcrumbData = [
    breadcrumbs.items[0],
    breadcrumbs.items[1],
    BUTTON_ELIPSIS,
    breadcrumbs.items[breadcrumbs.items.length -1]]
]
```

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,18 +1,58 @@
 <script lang="ts">
+  import {onMount} from "svelte";
+
   interface Breadcrumbs {
```
tajakobsen, line 4: If we want to be able to use this interface in the other application, move the interface to `.../ts/types.d.ts` so that it can be imported.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,18 +1,58 @@
...
+    ariaLabel?: string
+    showAllAriaLabel?: string
   }
 
   export let breadcrumbs: Breadcrumbs = {items: []};
+  export let loadJs = false
+
   let ariaLabel = breadcrumbs.ariaLabel ?? 'breadcrumbs';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'show all'
+  let spread = true
```
tajakobsen, line 15: Did you consider using xstate for state management?

tajakobsen (approved):
Good job! A few things to look at though.

In general try to use *css custom properties* where you can.

Consider adding [json-ld](https://developers.google.com/search/docs/advanced/structured-data/breadcrumb) to this component, so that it is always applied together with the markup.
Maybe [this npm can be useful for types for json-ld](https://github.com/google/schema-dts)

tajakobsen (changes requested):
Just changing state "Request changes"

src/svelte/Breadcrumbs.svelte
itemodahl: I feel it is a bit overkill.

src/scss/components/_buttons.scss
```diff
@@ -112,6 +112,21 @@ $button-colors: 'error' !default;
...
+    background: transparent;
+    color: var(--color);
+    text-decoration: underline;
+    min-height: 17px;
```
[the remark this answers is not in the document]
itemodahl: It is set to override the button default `min-height: 44px;`. The number comes from the sibling elements that are 17 px high. Changing it to auto.

src/scss/components/_buttons.scss
itemodahl: Lisa said we should follow the required wcag standard. Setting it to "auto".

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,21 +1,50 @@
...
 
   export let breadcrumbs: Breadcrumbs = {items: []};
-  let ariaLabel = breadcrumbs.ariaLabel ?? 'breadcrumbs';
+  export let loadJs = false
+
+  let ariaLabel = breadcrumbs.ariaLabel ?? 'brødsmulesti';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'vis mer'
+  let spread = true
+  let breadcrumbData = breadcrumbs.items
+  const BUTTON_ELLIPSIS = {url: '', title: '...', collapsed: true}
```
tajakobsen, line 12: I would still like to see a `Symbol()` used here.

tajakobsen (changes requested):
Looks good. I would still love to see a Symbol used as a placeholder for the elipsis in the array.

And maybe use of xstate.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,21 +1,99 @@
...
+      expandBreadcrumbs: (context) => {
+        context.breadcrumbsItems = breadcrumbs.items
+      }
+    }
+  })
+
+  const {state, send} = useMachine(breadcrumbsMachine)
+
+  let ariaLabel = breadcrumbs.ariaLabel ?? 'brødsmulesti';
+  let showAllBreadCrumbs = breadcrumbs.showAllAriaLabel ?? 'vis mer'
```
tajakobsen, line 68: Maybe use a name that clarifies that this is a string meant as a label?

```suggestion
  let showAllBreadCrumbsLabel = breadcrumbs.showAllAriaLabel ?? 'vis mer'
```

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,21 +1,99 @@
...
+  type BreadcrumbsState =
+    | {value: 'serverRendered'; context: BreadcrumbsContext}
+    | {value: 'closed'; context: BreadcrumbsContext}
+    | {value: 'open'; context: BreadcrumbsContext};
+
+  const breadcrumbsMachine = createMachine<BreadcrumbsContext, BreadcrumbsEvent, BreadcrumbsState>({
+    id: 'breadcrumbs',
+    initial: 'serverRendered',
+    context: {
+      breadcrumbsItems: []
```
tajakobsen, line 26: If you to to turn of JavaScript (using Storybook) you will see that breadcrumbs doesn't render.

You should initialize this array with `breadcrumbs.items`, instead of an empty `array`.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,21 +1,99 @@
...
+    },
+    states: {
+      serverRendered: {
+        on: {
+          MOUNTED: [
+            {
+              target: 'closed',
+              cond: (() => breadcrumbs.items.length > 3)
```
tajakobsen, line 34: If `context` was initialized with `breadcrumbs.items`, then you can probably use `context` as a parameter to the arrow function, and use it's value to do this check.

Also extract `3` into a named variable, so that someone reading this code can understand the meaning of `3`.

src/svelte/Breadcrumbs.svelte
```diff
@@ -1,21 +1,99 @@
...
-  <ol>
-    {#each breadcrumbs.items as item, index}
-      <li>
-        {#if index + 1 < breadcrumbs.items.length}
+  <ol class:expanded={isOpen}>
+    {#each bredcrumbs as item, index}
+      <li class:ellipsis={!isOpen && index +1 === bredcrumbs.length}>
+        {#if index === 0}
+          <a href={item.url}>Hjem</a>
+        {:else if item === BUTTON_ELLIPSIS && !onServer}
```
tajakobsen, line 85: Do we need `!onServer`? Because logically that can never be the case when `item === BUTTON_ELLIPSIS`

tajakobsen (changes requested):
Very good job! :+1: I just have a few small suggestions to look at.

tajakobsen (changes requested):
Good job. Maybe make the `3` a `const` instead of an `export let`.

tajakobsen (changes requested):
Prettier has not been run on at least Breadcrumbs.svelte

tajakobsen (approved):
Great job! Closing and merging to main.

PR #72, opened 2021-11-15
Responsive stepper

lha047:
Relates to MT-444

### DIFF
--- a/src/scss/components/_form-steps.scss
+++ b/src/scss/components/_form-steps.scss
@@ -17,11 +17,12 @@ ol.steps {
     display: flex;
     align-items: center;
     margin-top: 0;
+    font-size: $font-size-small;
 
     &:before {
       margin-right: 0;
       display: inline-flex;
-      border: 2px solid var(--color-mt-brown);
+      border: 2px dotted var(--color-mt-brown);
       counter-increment: step;
       content: counter(step);
       background-color: transparent;
@@ -43,6 +44,9 @@ ol.steps {
     &.steps__complete:before {
       color: var(--color-mt-background);
       background-color: var(--color-mt-brown);
+    }
+    &.steps__current:before {
+      border-style: solid;
     }
   }
 
--- a/src/scss/components/_form-steps.scss
+++ b/src/scss/components/_form-steps.scss
@@ -17,10 +17,10 @@ ol.steps {
     display: flex;
     align-items: center;
     margin-top: 0;
-    font-size: $font-size-small;
+    font-size: 0;
 
     &:before {
-      margin-right: $spacer-x-small;
+      margin-right: 0;
       display: inline-flex;
       border: 2px solid var(--color-mt-brown);
       counter-increment: step;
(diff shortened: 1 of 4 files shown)


stories/FormSteps.stories.svelte
```diff
@@ -15,15 +16,11 @@
...
-          <span class="inclusively-hidden">Ferdig: </span>
-        {/if}
-
-        {step}
-      </li>
-    {/each}
-  </ol>
+  <FormSteps
+    completedText="Utført"
+    currentText="Gjeldene side"
```
itemodahl, line 21: These two should be localized, with a fallback in Norwegian.

itemodahl (changes requested):
Very good! 

Only one change request: that we should be able to add localized phrases into the steps. With a Norwegian fallback.

stories/FormSteps.stories.svelte
lha047: Dont think its necessary in the story. (We dont have the localize function in this project)

src/svelte/FormSteps.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  export let labels: Array<string> = [];
+  export let completed: number = 0;
```
tajakobsen, line 3: Intellij suggests removing `:number` since it's easy to infer it here.

src/svelte/FormSteps.svelte
```diff
@@ -0,0 +1,28 @@
...
+  export let completedText = 'Ferdig';
+  export let currentText = 'Nåværende';
+  export let ariaValueText = `Steg ${completed + 1} av ${labels.length}`;
+</script>
+
+<ol
+  class="steps"
+  role="progressbar"
```
tajakobsen, line 11: I actually don't think you can add `role="progressbar"` to `<ol>`. You change the semantics of the element, and then the child `<li>` doesn't make sense.

src/svelte/FormSteps.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  export let labels: Array<string> = [];
+  export let completed: number = 0;
+  export let completedText = 'Ferdig';
+  export let currentText = 'Nåværende';
+  export let ariaValueText = `Steg ${completed + 1} av ${labels.length}`;
```
tajakobsen, line 6: It reads out which step is completed, when you would think it reads out which step you are currently on.
```suggestion
  export let ariaValueText = `Steg ${completed + 2} av ${labels.length}`;
```

tajakobsen: Does this need to have `export`?

src/svelte/FormSteps.svelte
```diff
@@ -0,0 +1,28 @@
...
+  aria-valuetext={ariaValueText}
+>
+  {#each labels as label, index (label)}
+    <li class:steps__complete={index <= completed}>
+      {#if index < completed}
+        <span class="inclusively-hidden">{completedText}: </span>
+      {/if}
+      {#if index === completed}
+        <span class="inclusively-hidden">{currentText}: </span>
+      {/if}
```
tajakobsen, line 24: Are you getting any of the `inclusively-hidden` texts in your screen reader? I'm only experiencing the `role="progressbar"`.

If they are not used, maybe we shouldn't have them?

src/scss/components/_form-steps.scss
```diff
@@ -17,10 +17,10 @@ ol.steps {
     display: flex;
     align-items: center;
     margin-top: 0;
-    font-size: $font-size-small;
+    font-size: 0;
```
tajakobsen, line 20: What is up here? `font-size: 0;`?

tajakobsen (changes requested):
Good job. But I have a few things to take a look at.

src/scss/components/_form-steps.scss
lha047: Its to make the text disappear on small screen safari. 
The alternative is to put a `<span class="responsive-hide">{label}</span>`. This is needed because of the `:before` and `:after`

tajakobsen: I have never seen this done before.

I see two options here.
 1. Change the code to add a class like you write above
 2. Add a comment here to clarify what is happening. Because this might confuse other future readers that haven't seen this before either.

tajakobsen (changes requested):
Good job, we are getting close. You dodn't address the comment I had before about `ariaValueText`.

And I think we need something to make that `font-size: 0` more clear to readers.

src/svelte/FormSteps.svelte
lha047: I think it should be so it can be translated

lha047:
I think we should keep the `export` on `ariaValueText ` for translation

tajakobsen (approved):
Looks good. Closing and merging to main

PR #73, opened 2021-11-16
Remove initial accordion animation

tajakobsen:
Also remove arrow icon when header is not a button, because it isn't useful in non-javascript environments.

Relates to: MT-552


lha047 (approved):
Nice job :D

PR #85, opened 2021-11-22
Mt 338 grid layout

lha047:
Add grid layout

Relates to MT-338


tajakobsen (changes requested):
It looks good. But I think we should go in a direction where we use some general utility classes to create the grids, instead of using selectors that exists in the templates.

src/svelte/Breadcrumbs.svelte
```diff
@@ -83,7 +84,7 @@
   }
 </script>
 
-<nav class="breadcrumbs" aria-label={ariaLabel}>
+<nav class={`breadcrumbs ${classNames}`} aria-label={ariaLabel}>
```
tajakobsen, line 87: Would this be the same?

```suggestion
<nav class="breadcrumbs {classNames}" aria-label={ariaLabel}>
```

src/svelte/Breadcrumbs.svelte
```diff
@@ -7,6 +7,7 @@
   const BUTTON_ELLIPSIS = Symbol();
   export let breadcrumbs: Breadcrumbs = {items: []};
   export let loadJs = true;
+  export let classNames = '';
```
tajakobsen, line 10: How about letting the attribute be named `class` externally?

```suggestion
  let classNames = '';
  export {classNames as class};
```

tajakobsen (changes requested):
Very good1 I think we are heading in the right direction here :)

Just a few minor comments to look at.

tajakobsen (approved):
Looks good. Closing and merging to main.

PR #102, opened 2021-11-30
Add configurable classes to Disclosure

tajakobsen:
Relates to: MT-388

### DIFF
--- a/src/svelte/Disclosure.svelte
+++ b/src/svelte/Disclosure.svelte
@@ -12,6 +12,10 @@
   export let title: string;
   export let headerTag: 'h2' | 'h3' | 'h4' = 'h3';
   export let theme: 'bordered' | 'links' = 'bordered';
+  export let headerClass: string | undefined;
+  export let panelClass: string | undefined;
+  let disclosureClass: string | undefined;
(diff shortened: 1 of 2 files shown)


src/svelte/Disclosure.svelte
```diff
@@ -12,6 +12,10 @@
   export let title: string;
   export let headerTag: 'h2' | 'h3' | 'h4' = 'h3';
   export let theme: 'bordered' | 'links' = 'bordered';
+  export let headerClass: string | undefined;
```
lha047, line 15: Is there a difference between 
```
export let headerClass: string | undefined;
{panelClass || ''}
```
and
```
export let headerClass: string = '';
{panelClass}
```

Seems easer to read with default value?

lha047 (changes requested):
Good job :D 
Just have a question about using a default value.

src/svelte/Disclosure.svelte
tajakobsen: I don't have any strong opinions either way, so I can change it.

lha047 (approved):
Nice :D

PR #103, opened 2021-12-01
MT-546 Feat Add hover style for bordered accordion

lha047:
- bold and indentation for hover
- fix font-size bordered accordion

Relates to MT-546

### DIFF
--- a/src/scss/components/_disclosure.scss
+++ b/src/scss/components/_disclosure.scss
@@ -40,7 +40,6 @@
 
       &:hover {
         padding-left: var(--spacer-xx-small);
-        font-weight: $font-weight-bold;
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/scss/components/_disclosure.scss
```diff
@@ -36,6 +36,12 @@
 
     .disclosure-header {
       justify-content: space-between;
+      transition: padding-left ease-in-out 150ms;
+
+      &:hover {
+        padding-left: var(--spacer-xx-small);
+        font-weight: $font-weight-bold;
```
tajakobsen, line 43: I don't like how the `font-weight` change feels. We should have a talk with @martevb about this.

tajakobsen (changes requested):
Good job. But I think we should consider removing the bold effect on hover. It feels too heavy handed.

PR #109, opened 2021-12-01
MT-553 Change hard coded phrase (x tegn igjen) to dynamic

itemodahl:
MT-553

### DIFF
--- a/src/svelte/form/TextArea.svelte
+++ b/src/svelte/form/TextArea.svelte
@@ -7,6 +7,7 @@
   export let value;
   export let name: string;
   export let label: string;
+  export let countCharactersLeftLabel: string;
   export let error: ErrorDetail | undefined;
   export let helpText: string | undefined;
   export let textOptional: string | undefined;
@@ -36,7 +37,7 @@
 <textarea
   id={name}
   {name}
-  use:countCharacters
+  use:countCharacters={countCharactersLeftLabel}
   class="form-field"
   bind:value
   {maxlength}
--- a/src/svelte/form/TextArea.svelte
+++ b/src/svelte/form/TextArea.svelte
@@ -37,7 +37,7 @@
 <textarea
   id={name}
   {name}
-  use:countCharacters={countCharactersLeftLabel}
+  use:countCharacters={countCharactersLeftLabel ? {countCharactersLeftLabel} : undefined}
   class="form-field"
   bind:value
   {maxlength}
--- a/src/svelte/form/TextArea.svelte
+++ b/src/svelte/form/TextArea.svelte
@@ -7,7 +7,7 @@
   export let value;
   export let name: string;
   export let label: string;
-  export let countCharactersLeftLabel: string;
+  export let countCharactersLeftLabel: string | undefined;
   export let error: ErrorDetail | undefined;
   export let helpText: string | undefined;
   export let textOptional: string | undefined;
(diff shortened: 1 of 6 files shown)


src/svelte/form/TextArea.svelte
```diff
@@ -7,6 +7,7 @@
   export let value;
   export let name: string;
   export let label: string;
+  export let countCharactersLeftLabel: string;
```
tajakobsen, line 10: We still allows `undefined` as a legal value here, since we provide a default value in  `countCharacters`, right?

```suggestion
  export let countCharactersLeftLabel: string | undefined;
```

src/ts/count-characters.ts
```diff
@@ -1,12 +1,15 @@
 import type {Action} from './types';
 
-export const countCharacters: Action<HTMLInputElement | HTMLTextAreaElement> = node => {
+export const countCharacters: Action<HTMLInputElement | HTMLTextAreaElement> = (
+  node,
+  countCharactersLeftLabel: string = 'tegn igjen'
```
tajakobsen, line 5: Since we can only have one parameter as input to an action, I think it would be most future proof to let it be an object, with the value inside it. That way it can be expanded with more parameters later.

This same principle goes for custom events too.

```suggestion
  parameters: { countCharactersLeftLabel: string } = {countCharactersLeftLabel: 'tegn igjen'}
```

tajakobsen (changes requested):
Looks good. I just have a few small tings that can be good to consider.

src/ts/count-characters.ts
itemodahl: Agree. Changing it now.

src/svelte/form/TextArea.svelte
```diff
@@ -36,7 +37,7 @@
 <textarea
   id={name}
   {name}
-  use:countCharacters
+  use:countCharacters={countCharactersLeftLabel}
```
tajakobsen, line 40: You are not wrapping the string in an object, while `countCharacters` expects an object.

src/ts/count-characters.ts
```diff
@@ -1,12 +1,15 @@
 import type {Action} from './types';
 
-export const countCharacters: Action<HTMLInputElement | HTMLTextAreaElement> = node => {
+export const countCharacters: Action<HTMLInputElement | HTMLTextAreaElement> = (
+  node,
+  params: {countCharactersLeftLabel: string} = {countCharactersLeftLabel: 'tegn igjen'}
```
tajakobsen, line 5: Use "TextArea with error" to check the fallback case (`'tegn igjen'`).

tajakobsen (changes requested):
The preview in storybook is broken. I have suggested what in a comment what the fix is.

Also `countCharacters` is also used by `TextInput.svelte`, so it should also be updated with this change.

stories/components/form/TextArea.stories.svelte
```diff
@@ -11,23 +11,26 @@
     helpText:
       'Beskriv kort og konkret hva du har observert og hvor alvorlig hendelsen er. Vær oppmerksom på den ansvarlige ofte får se meldingen.',
     errorMessage: 'Fyll inn dette feltet.',
-    disableCss: false
+    disableCss: false,
+    countCharactersLeftLabel: 'Karakterer igjen'
```
lha047, line 15:
```suggestion
    countCharactersLeftLabel: 'karakterer igjen'
```

tajakobsen (approved):
Looks good! Good job.

PR #110, opened 2021-12-01
Remove styles from "_theme.scss" to prevent re-import

tajakobsen:
Also add `.fleat-right` utility class.

Relates to: MT-606


lha047 (approved):
LGTM 🎄

PR #124, opened 2021-12-07
Fix reactivity on breadcrumbs

tajakobsen:
Relates to: MT-610


lha047 (changes requested):
Looks the code works with regards to updating the breadcrumbs between pages. But the `...` for long breadcrumbs is not working any more.

Screenshot 2021-12-07 at 15 32 09
e.

lha047 (approved):
Looks good to me :D

PR #139, opened 2021-12-13
MT-646:Add rel=external to breadcrumbs

lha047:
- to make proxy hadle navigation on breadcrumbs

Relates to MT-546


itemodahl (approved):
Well done

PR #141, opened 2021-12-13
MT-376: Update published style

lha047:
- for moved under intro

MT-376


itemodahl (changes requested):
You might want to update the story for article to show the date after the preface.

itemodahl (approved):

PR #148, opened 2021-12-17
MT-650: hightlighted content

lha047:
Add component and styling for highlighted content

MT-650

### DIFF
--- a/src/svelte/HighlightedContentLink.svelte
+++ b/src/svelte/HighlightedContentLink.svelte
@@ -16,7 +16,7 @@
   rel={isExternal ? 'external' : undefined}
   data-testid="hightlighted-content-link"
 >
-  {#if image?.src}
+  {#if image && image.src}
     <img src={image.src} alt={image.alt} />
   {/if}
   <div class="layout-flex layout-flex-col highlighted-text" style="gap:0">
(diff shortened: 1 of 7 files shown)


src/svelte/HighlightedContentLink.svelte
```diff
@@ -0,0 +1,30 @@
...
+  {/if}
+  <div class="layout-flex layout-flex-col highlighted-text" style="gap:0">
+    {#if heading}
+      <h3 class={`heading ${headingClass}`}>
+        {heading}
+      </h3>
+    {/if}
+    <slot />
+  </div>
+</a>
```
itemodahl, line 30: I don't think `div` is a valid content in an anchor element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a . Should we use `button` instead?

src/ts/utils.ts
```diff
@@ -10,3 +10,7 @@ export function toKebabCase(text: string): string {
     .replace(/ /g, '-')
     .toLowerCase();
 }
+export function displayDataTime(lang: string, isoDate?: string): string {
+  if (!isoDate) return '';
+  return new Intl.DateTimeFormat(lang).format(new Date(isoDate));
+}
```
itemodahl, line 16: Should the check if isoDate exists happen outside this function. And instead make isoDate mandatory argument?

itemodahl (changes requested):
Good job! A couple of feedbacks you can take a look at.

src/svelte/HighlightedContentLink.svelte
lha047: Looks to me like you should not use interactive elements either inside link.

PR #154, opened 2021-12-17
Fix deprecated static-dir flag waring

lha047:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag


itemodahl (approved):
Good

PR #229, opened 2022-02-01
MT-676: add support for html encoded characters

itemodahl:
MT-676


lha047 (approved):
Nice :D 

Consider making tests next time ;)

PR #233, opened 2022-02-03
MT-733: Add styling for legal text disclosure

lha047:
MT-733

### DIFF
--- a/src/svelte/Disclosure.svelte
+++ b/src/svelte/Disclosure.svelte
@@ -84,7 +84,7 @@
   {#if isOpen}
     <div
       id={bodyId}
-      class="disclosure-panel {panelClass}"
+      class="disclosure-panel {panelClass} {onServer ? 'on-server' : ''}"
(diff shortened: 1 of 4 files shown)


src/svelte/Disclosure.svelte
```diff
@@ -11,7 +11,7 @@
   export let loadJs = true
   export let title: string
   export let headerTag: 'h2' | 'h3' | 'h4' = 'h3'
-  export let theme: 'bordered' | 'links' = 'bordered'
+  export let theme: 'bordered' | 'links' | 'legal-text' = 'bordered'
```
tajakobsen, line 14: Is `legal-text` a good name here? The other two legal values are agnostic for where they are used. If we want to use this Disclosure somewhere different then "legal texts", then having an agnostic name would be better.

tajakobsen (changes requested):
I'm not able to see the svg-icons in the preview (Neither in Chrome nor Firefox). It just looks like normal <h2> elements.

tajakobsen (approved):
Looks good. Closing and merging to main.

PR #243, opened 2022-02-14
Created mixin article-spacing for styles in the article-page class

miloszhygen:
and reused the mixin inside legal-collection class


lha047 (approved):
Nice :D

PR #265, opened 2022-02-23
💄 MT-891  Moved style out of the article-spacing mixin

miloszhygen:
the summary details needs less space on top and bottom.

### DIFF
--- a/src/scss/components/_legal-collection.scss
+++ b/src/scss/components/_legal-collection.scss
@@ -4,12 +4,16 @@
 .legal-collection {
   @include article-spacing;
 
+  section > details,
+  section + .disclosure {
+    margin-top: var(--spacer-medium);
+  }
   .intro {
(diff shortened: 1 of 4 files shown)


src/scss/components/_legal-collection.scss
```diff
@@ -4,12 +4,16 @@
 .legal-collection {
```
lha047, line 4: Can you add a `LegalGuidanceCollection.stories.svelte` to display how the legal guidance collection should look like?

lha047 (changes requested):
Code looks good :D 

Could you make a story for legal guidance collection?

src/scss/components/_legal-collection.scss
miloszhygen: I'll fix :)

lha047 (approved):
Looks good :D

PR #268, opened 2022-02-24
💄 MT-828: Set max length for better readablility

itemodahl:
MT-828

### DIFF
--- a/src/scss/components/_top-tasks.scss
+++ b/src/scss/components/_top-tasks.scss
@@ -1,10 +1,6 @@
 @use '../theme' as *;
 
 .top-tasks-container {
-  h2,
-  .top-tasks-text {
-    max-width: 760px;
-  }
   h2 + .top-tasks-text {
     margin-top: var(--spacer-x-small);
   }
--- a/src/scss/components/_top-tasks.scss
+++ b/src/scss/components/_top-tasks.scss
@@ -1,6 +1,10 @@
 @use '../theme' as *;
 
 .top-tasks-container {
+  h2,
+  .top-tasks-text {
+    max-width: 760px;
+  }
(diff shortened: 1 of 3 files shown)


src/scss/components/_top-tasks.scss
```diff
@@ -1,6 +1,10 @@
 @use '../theme' as *;
 
 .top-tasks-container {
+  h2,
```
lha047, line 4: This should be solved using the grid system in the web project.
From figma it looks like there is a need for `col-1-span-7` in the `_grid.scss`. And that should be used on the heading and text in the web project.
Just ask if anything is unclear :)

lha047 (changes requested):
This should be solved using grid columns. See my comments :)

stories/components/content/TopTasks.stories.svelte
```diff
@@ -38,27 +49,29 @@
...
-        class:layout-flex--basis-2={index !== 0 && index !== 1 && index !== 3}
-        aria-labelledby={'toptask'}
-        style="gap: var(--spacer-x-small);">
-        {#each topTasks.slice(0, 6 - index) as link}
-          <a href={link.href} class="button button--primary top-task">
-            <h3 class="heading">{@html link.heading}</h3>
-            <span class="text">{link.text}</span>
+<Story name="Normal" let:topTasks let:icon let:disableCss let:title let:intro let:path>
+  <div class="layout-grid layout-grid--column-12 top-tasks-container">
+    <h2 id={path} class="top-tasks-text col-1-span-12">
```
lha047, line 54: Think this should be 7 columns too?

lha047 (approved):
Nice work :D

Just one more small detail. You can merge when its fixed :)

PR #305, opened 2022-03-22
MT-731 Removed underline when hovering a link

miloszhygen:
- remove underline on links
- remover underline in linklist AND add indent

[MT-731](https://itemtest.atlassian.net/browse/MT-731)


lha047 (approved):
Looks good to me :D

PR #308, opened 2022-03-23
♿ MT-499: Add aria-describedby and aria-live

lha047:
- to inform screenreaders about characters left

MT-499


miloszhygen (approved):
Works like a charm 🎉  It would be easier to see changes to logic if formatting and logic changes was done into two different commits?

PR #335, opened 2022-04-04
💄MT-1009 adjust breadcrumbs mobile

lha047:
On mobile: 
- Only previous breadcrumb should be displayed. Arrow points left. Smaller spacing between arrow and text.

MT-1009, MT-945


itemodahl (approved):
Good job!

PR #347, opened 2022-04-07
Fix dont show group when link list is empty

lha047:
MT-664


itemodahl (approved):
Very good

lha047:
Already done

PR #370, opened 2022-04-19
💄MT-763 table of contents

itemodahl:
MT-763

### DIFF
--- a/src/scss/global/_links.scss
+++ b/src/scss/global/_links.scss
@@ -38,6 +38,19 @@ a {
     }
   }
 
+  &.down-arrow {
+    display: flex;
+    align-items: center;
+    color: var(--color-links);
+    &::before {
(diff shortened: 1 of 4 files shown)


src/scss/global/_links.scss
```diff
@@ -38,6 +38,19 @@ a {
     }
   }
 
+  &.down-arrow {
+    display: flex;
+    align-items: center;
+    color: var(--color-links);
```
lha047, line 44: Dont think this is needed

src/scss/global/_links.scss
```diff
@@ -38,6 +38,19 @@ a {
...
+    &::before {
+      content: url('data:image/svg+xml,%3Csvg%20width%3D%2214%22%20height%3D%2214%22%20fill%3D%22none%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%3Cpath%20d%3D%22M7.002.11%205.788%201.325l4.805%204.813H.113V7.86h10.48l-4.805%204.814%201.214%201.214L13.892%207%207.001.11Z%22%20fill%3D%22%23075B7A%22%2F%3E%3C%2Fsvg%3E');
+      transform: rotate(90deg);
```
lha047, line 47: Think it would be better to use an arrow that points down

lha047 (changes requested):
Nice :D 

Left two comments

lha047 (approved):
Nice :D

PR #376, opened 2022-04-22
♿️ MT-909 fix cypress a11y

lha047:
Fix a11y issues

Mt-909


miloszhygen (approved):
All cypress and jest tests pass 🚀

PR #392, opened 2022-05-11
Mt 969 header spacing

miloszhygen:
This PR fixes:

image

[MT-969 ](https://itemtest.atlassian.net/browse/MT-969)


src/scss/utilities/_layout.scss
```diff
@@ -178,12 +178,17 @@ Only use if it does NOT affect accessibility
...
+header{
+  border-bottom: $border-red-regular;
+}
+
 .header {
   display: grid;
   grid-template-columns: 1fr min-content min-content min-content;
   align-items: center;
-  border-bottom: $border-red-regular;
+  width: min(90rem, 95vw);
```
lha047, line 189: Maybe use the already existing class `container` for on the html over adding the same to the the `.header`-class

lha047 (approved):
Looking good :D

Have just one refac suggestion

PR #407, opened 2022-05-23
💄 MT-763 table of contents margin

itemodahl:
MT-763

### DIFF
--- a/src/scss/utilities/_mixins.scss
+++ b/src/scss/utilities/_mixins.scss
@@ -175,9 +175,12 @@
       margin-top: var(--space-section);
     }
     &.table-of-contents {
-      margin: 4.06rem 0 2.81rem;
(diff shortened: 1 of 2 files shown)


src/scss/utilities/_mixins.scss
```diff
@@ -175,7 +175,7 @@
       margin-top: var(--space-section);
     }
     &.table-of-contents {
-      margin: var(--spacer-large) 0 3.25rem;
+      margin: 4.06rem 0 2.81rem;
```
lha047, line 178: I dont think we should use specific sizes for margins. Try adjusting surrounding margins. We are trying to follow the principle of working with top-margins on the elements.

lha047 (changes requested):
Nice

See my comment

lha047 (approved):
Good job :D

PR #421, opened 2022-05-24
💄 MT-963 Add styling for next previous links

lha047:


src/scss/utilities/_layout.scss
```diff
@@ -49,9 +49,21 @@
...
   }
+  &--no-wrap {
+    flex-wrap: nowrap;
+  }
+  &--align-self-end {
+    align-self: flex-end;
+  }
+  &--row-reverse {
+    flex-direction: row-reverse;
+  }
```
itemodahl, line 66: I know it can be practical to mirror css-rules, but I think we should be careful doing it. We are moving the css-styling to the html code and it is kinda the same as doing inline styling `style="flex-direction: row-revers"` . What do you think?

PR #425, opened 2022-05-25
💄MT-1147: increase margin on blue accordions

itemodahl:
MT-1147


lha047 (changes requested):
This should maybe be special treatment of the lovdata texts?

PR #429, opened 2022-05-31
MT-1168 Fixed an issue where sub-menu items did not allign to parent …

miloszhygen:
…title text

[MT-1168](https://itemtest.atlassian.net/browse/MT-1168)


src/scss/components/_disclosure.scss
```diff
@@ -101,7 +101,7 @@
 
   &.disclosure-no-border {
     > .disclosure-header {
-      gap: var(--spacer-xx-small);
+      gap: var(--spacer-xxx-small);
```
lha047, line 104: This also affects the other `.disclosure-no-border`. Dont think it should do that. Maybe it should be moved to menu styling?

src/scss/components/_disclosure.scss
```diff
@@ -135,7 +135,12 @@
     }
     .disclosure-panel {
       border-top: $border-width-small solid var(--color-secondary);
-      margin: 0 var(--spacer-x-small);
+      margin: 0 var(--spacer-xxx-small);
```
lha047, line 138: This also affects the other `.disclosure-no-border`. Dont think it should do that. Maybe it should be moved to menu styling?

lha047 (changes requested):
Looks like this affects more than it should. 
I think it should be moved to menu styling. Lets talk about it.

src/scss/components/_disclosure.scss
miloszhygen: Fixing :)

src/scss/components/_disclosure.scss
miloszhygen: Fixing :)

lha047 (approved):
Nice :D

PR #431, opened 2022-06-01
♿️ MT-1163 Fix a11y

lha047:
- move link in side list. For screen reader consistency
- MT-1213 increase font-size menu level 2 link
- add aria-current to current page link.

### DIFF
--- a/src/svelte/content/MenuItems.svelte
+++ b/src/svelte/content/MenuItems.svelte
@@ -31,13 +31,16 @@
           headerClass="p-tb-xxs"
           {loadJs}
           icon={item.iconResource}>
-          <a
-            href={item.url}
-            rel={mapRelExternal(item.url)}
-            class="forward-arrow-small on-primary no-underline hover-indent {item.isActive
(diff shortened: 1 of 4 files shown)


src/svelte/content/MenuItems.svelte
```diff
@@ -30,19 +30,23 @@
...
+              <a
+                href={item.url}
+                class="forward-arrow-small on-primary no-underline hover-indent {item.isActive
+                  ? 'is-active'
+                  : ''}"
+                aria-current={item.isActive ? 'page' : false}>{item.title}</a>
+            </li>
             {#each item.children || [] as child}
-              <li>
+              <li class="menu-level-3">
```
miloszhygen, line 43: the `menu-level-3` class has no other function than to tell what level you're on? There is no styling on this class?

lha047: No functionality. Removing

miloszhygen (approved):
🎉

PR #432, opened 2022-06-01
💄MT-1169 menu styling mobil

lha047:
- fix sizing mobile


src/scss/components/_dropdown.scss
```diff
@@ -269,28 +277,49 @@
   }
 }
 .menu {
+  > li {
```
itemodahl, line 280: `>` means "all `li` that has `.menu` as parent". Wouldn't `.menu li` be sufficient here?

itemodahl (approved):
Good job!

src/scss/components/_dropdown.scss
lha047: No it means direct child of menu. I dont want other sublists to be affected

PR #434, opened 2022-06-02
🐛 MT-1161 fix external links

lha047:
MT-1161


miloszhygen (approved):
Works like a charm! ✨

PR #437, opened 2022-06-07
✨MT-799 tags component

lha047:
MT-799


src/svelte/Tags.svelte
```diff
@@ -0,0 +1,29 @@
+<script lang="ts">
+  export let tags: Array<{text: string; ariaLabel: string}> = []
+  export let isClosable = false
+
+  function handleClick(index: number) {
+    tags = tags.filter((tag, ind) => {
```
miloszhygen, line 6: what about calling it `tagIndex` instead of `ind`?

lha047: You got it :D

miloszhygen (approved):
Nice 🎉

PR #441, opened 2022-06-08
💄MT-1179 sub section menu 2 col

lha047:
MT-1179


stories/Header.stories.svelte
```diff
@@ -47,7 +47,8 @@
           iconResource: drinkIcon,
           children: [
             {title: 'Produksjon av mat', url: 'https://www.mattilsynet.no/mat_produksjon'},
-            {title: 'Matservering', url: '/mat/servering'}
+            {title: 'Matservering', url: '/mat/servering'},
+            {title: 'Lokalmat', url: '/mat/lokalmat'}
           ]
```
miloszhygen, line 52: should we use `#` as `url`? Then we dont navigate to an empty page with this error message `Cannot GET /mat/servering` ;)

miloszhygen, line 52: should we use `#` as `url`? Then we dont navigate to an empty page with this error message `Cannot GET /mat/servering` ;)

miloszhygen (approved):
🙌

PR #458, opened 2022-06-22
💄 MT-1190 Fix styling menu right

lha047:
- remove flex styling from menu item
- update forward arrow end style
- update header story to show fix


itemodahl (approved):
LGTM

PR #463, opened 2022-06-24
💄MT-1232 menu mobile styling

lha047:


src/scss/components/_dropdown.scss
```diff
@@ -319,7 +333,7 @@
     }
     & > li:last-child {
       border-bottom: $border-width-small solid var(--color-mt-lighter-light-blue);
-      padding-bottom: 1px;
+      padding-bottom: 3px;
       margin-bottom: 0;
```
itemodahl, line 337: Should you use rem here?

itemodahl (approved):
LGTM

src/scss/components/_dropdown.scss
lha047: Fixed

PR #483, opened 2022-08-01
✨ MT-790 multiselect

lha047:


itemodahl (changes requested):
Very good job! Just a couple of remarks.

The tag components circle 'x' should be inverted instead of just disappearing.
And there should be a way to set a default text, which doesn't have a value. Like "Choose a country".

itemodahl (approved):

PR #526, opened 2022-09-21
Fix: MT-1135, MT-1329 guide menu, xl screens

lha047:


itemodahl:
The menu looks off in the design system.
Skjermbilde 2022-09-21 kl  12 07 08

itemodahl (approved):
Very good job!

PR #591, opened 2022-10-19
MT-1458: Do not return empty string

itemodahl:
MT-1458


lha047:
This is already merged in other PR

PR #610, opened 2022-10-28
MTP-486 adds css for banner image. With conditional breakpoint

amundlindstol:
image


src/scss/_theme.scss
```diff
@@ -36,6 +36,8 @@ $font-size-label: clamp(1.1rem, 4.333vw + 1rem, 1.5rem) !default; //24px
 $height-header-regular: 6.25rem;
 $height-header-mobile: 3.75rem;
 
+$height-banner: 28rem; // 448px
```
lha047, line 39: What is this height? Ask Camille if it is ok. 
In my opinion it looks like it takes too much space.

lha047 (approved):
Nice :) 

Check with Camille about the height of the image before merging

PR #613, opened 2022-11-02
MTP-1079 added a border radius class

amundlindstol:


src/scss/global/_utils.scss
```diff
@@ -78,3 +78,6 @@
 .border-radius-0 {
   border-radius: 0;
 }
+.border-radius {
+  border-radius: $border-radius;
```
lha047, line 82: What do we need this for?

amundlindstol: to expose the border-radius without using inline style. is there another way to reference it?
image

lha047 (approved):
Nice :D

PR #651, opened 2022-11-30
Fix: MTP-1189 fix table collection

lha047:
Checkout branch and test that the webpack5 dep does not break anything


amundlindstol (approved):
🥇 noice!

PR #675, opened 2022-12-15
MTP-484_current_issues

amundlindstol:
MTP-484


src/svelte/content/CurrentIssues.svelte
```diff
@@ -0,0 +1,43 @@
...
+      fill="#3F3E3E" />
```
lha047, line 24: Use a color from the color system?

lha047 (changes requested):
Left some comments

src/svelte/content/CurrentIssues.svelte
```diff
@@ -0,0 +1,53 @@
...
+    return index % 2 === 0 ? 'col-2-span-4' : 'col-7-span-5'
+  }
+</script>
+
+<section
+  class="layout-grid layout-grid--column-12 current-issues"
+  data-testid="current-issues"
+  aria-labelledby="current-issues-heading">
+  <h2 id="current-issues-heading" class="inclusively-hidden">{ariaLabel}</h2>
+  <svg
```
lha047, line 23: Just wondering if we need all this svg in the dom? 
Would it be better to have it as url-encoded css?

lha047 (approved):
Looks good to me :D 

Have a look at the comment

PR #698, opened 2023-01-02
MTP-1214: fix remove underline on hover menu

lha047:
- only apply hover underline to button link (menu trigger button)


amundlindstol (approved):
👍

PR #704, opened 2023-01-06
MTP-1196_fix_icon_placement_on_menu_points

amundlindstol:
MTP-1196


src/scss/components/_menu-points.scss
```diff
@@ -12,18 +12,22 @@
...
 
-  &--title {
-    grid-area: heading;
-    align-self: center;
-  }
-  .svg,
+  .svg-wrapper,
   img {
     grid-area: icon;
     align-self: center;
```
lha047, line 18:
```suggestion
    align-self: start;
```
Should align to top. Check with design that placing is ok

src/svelte/content/MenuPointsIcons.svelte
```diff
@@ -14,11 +14,13 @@
...
     <li class="menu-point">
       {#if menuPoint.iconResource}
-        <div class="svg" aria-hidden="true">
+        <div class="svg-wrapper" aria-hidden="true">
           {@html menuPoint.iconResource}
         </div>
       {:else if menuPoint.icon}
         <img src={menuPoint.icon} alt="" aria-hidden="true" data-testid="img-icon" />
+      {:else}
+        <div>&#8203;</div>
```
lha047, line 23: You dont need an empty div here to fill the space. Just place the link in the heading grid-area

lha047 (changes requested):
Left some comments :)

lha047 (approved):
Nice :D

PR #716, opened 2023-01-20
MTP-39_list_of_lokaliteter

amundlindstol:
MTP-39


lha047 (approved):
Nice :D
Get at design review from Camille.
Im ok with the way it is (for this thing that should not look great in the first place)

Have a look at the typing comment.

PR #725, opened 2023-02-03
MTP-452_last_page_should_not_be_a_link

amundlindstol:
MTP-452

### DIFF
--- a/src/svelte/components/Breadcrumbs.svelte
+++ b/src/svelte/components/Breadcrumbs.svelte
@@ -112,6 +112,10 @@
       <li class:ellipsis={!isFull && index + 1 > 2}>
         {#if index === 0}
           <a href={item.url} rel="external">{homeLabel}</a>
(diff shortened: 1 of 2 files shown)


src/svelte/components/Breadcrumbs.svelte
```diff
@@ -112,6 +112,10 @@
       <li class:ellipsis={!isFull && index + 1 > 2}>
         {#if index === 0}
           <a href={item.url} rel="external">{homeLabel}</a>
+        {:else if index + 1 === $state.context.breadcrumbsItems.length}
```
lha047, line 115: I think the last breadcrumb is in the else clause. 
The `aria-current="page"` has no effect on `span` element. You can drop it.

lha047 (changes requested):
Nice :D 
You can use the element in the `else` block

lha047 (approved):
Nice :D

PR #757, opened 2023-02-22
MTP-385: Add aria-details

itemodahl:
MTP-385


src/svelte/components/FileUploadButton.svelte
```diff
@@ -129,6 +129,7 @@
   class:error
   class:inclusively-hidden-fixed={!onServer}
   aria-describedby={createInputAriaDescribedby(name, error)}
+  aria-details="{name}-details"
```
lha047, line 132: This need to some how check that there is an id with `{name}-details`. Its bad practice to have these attributes pointing to nowhere.
Maybe the details should be a part of this component?

lha047 (changes requested):
Maybe the details should be a part of this component? See comment

PR #772, opened 2023-03-13
MTP-1351-vise-antall-treff

amundlindstol:


src/svelte/components/form/Search.svelte
```diff
@@ -3,19 +3,22 @@
...
   export let placeholder: string | undefined
+  export let ariaControls: string | undefined
   export let inputClass = ''
   export let primaryButton = false
 </script>
 
-<Label class={labelClass} for={name} showOptionalText={false}>
-  {label}
-</Label>
+{#if label}
```
lha047, line 17: If we are not going have a visible label. We need to have another way to explain what this is for accessibility

src/svelte/components/form/Search.svelte
```diff
@@ -3,19 +3,22 @@
...
-  export let name: string
-  export let label: string
-  export let labelClass: string
+  export let name: string | undefined
+  export let label: string | undefined
+  export let labelClass: string | undefined
   export let helpText: string | undefined
   export let searchButtonText: string | undefined
   export let placeholder: string | undefined
+  export let ariaControls: string | undefined
```
lha047, line 12: Where is the thing that the button controls?

lha047 (changes requested):
I think we need to have an extra look at this with regards to accessibility

src/svelte/components/form/Search.svelte
amundlindstol: https://github.com/Mattilsynet/mattilsynet-web/pull/358/files#diff-bfa4d0f215637200d8d16f37f6040e0f0d048cfde50ee1a51d34bb83601283a1R35-R40

src/svelte/components/form/Search.svelte
amundlindstol: VoiceOver says: "search text field, search".
I feel like it gives the same amount of information as the design
image

src/svelte/components/form/Search.svelte
```diff
@@ -37,6 +40,7 @@
     <button
       type="submit"
       class="button button--flat form-field"
+      {ariaControls}
```
lha047, line 43: This will not produce the right attribute. It is currently `ariacontrols` and needs to be `aria-controls`

lha047 (changes requested):
Look at aria controls

lha047 (approved):
Good job :D

PR #794, opened 2023-04-04
MTP-1473 focus remove search

lha047:
- fix focus on items in header
- fix focus on breadcrumbs
- fix search outside of view on mobile

### DIFF
--- a/src/svelte/components/form/Search.svelte
+++ b/src/svelte/components/form/Search.svelte
@@ -15,7 +15,6 @@
   export let inputClass = ''
   export let loadJs = false
 
-  $: hasValue = value && value !== ''
   onMount(() => {
     loadJs = true
   })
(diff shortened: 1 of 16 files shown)


src/svelte/components/form/Search.svelte
```diff
@@ -10,8 +11,16 @@
   export let searchButtonText: string | undefined
   export let placeholder: string | undefined
   export let ariaControls: string | undefined
+  export let ariaRemoveTextLabel = 'Tøm'
   export let inputClass = ''
-  export let primaryButton = false
+  let isInitialized = false
+
+  beforeUpdate(() => {
+    if ((value === undefined || value === '') && !isInitialized && document) {
```
amundlindstol, line 19: (value === undefined || value === '') -> !value
... but i think you can remove the check entirely? isInitialized should be enough

amundlindstol: also... can document be undefined?

amundlindstol (approved):
LGMT :D
See comments :)

src/svelte/components/form/Search.svelte
petterkoding: What does beforeUpdate do in this situation?

Isn't it possible to not use it at all and just check {#if value} or {#if value.length > 0} on line 49?

petterkoding (approved):

PR #810, opened 2023-04-12
MTP-354 Add appeal-decision result icons

petterkoding:
Icons used for the results in appeal decision 🟢🔴🟠

Success (green), Warning (orange) , Error(red)


itemodahl (approved):
Gj

PR #813, opened 2023-04-14
Use common css

lha047:


.storybook/preview.js
```diff
@@ -1,4 +1,4 @@
-import '../dist/app.css';
+import '@mattilsynet/designsystem-css/dist/app.css'
```
TheBoosja, line 1: Is it possible to remove `dist` from this. Ideally, I think it should be something like:
```suggestion
import '@mattilsynet/designsystem-css/app.css'
```

mt-kit/core/css/package.json
```diff
@@ -1,8 +1,22 @@
...
+    "Mattilsynet",
+    "CSS"
+  ],
   "files": [
-    "dist"
+    "dist",
+    "src/_theme.scss",
+    "src/_color-tokens.scss",
+    "src/utilities/_mixins.scss",
+    "src/utilities/colors.js"
```
TheBoosja, line 19: I think we should only expose what we must. If it is possible to just publish `app.css` and the files under `src`, so that practically it becomes:
```
import '@mattilsynet/designsystem-css/app.css'
import '@mattilsynet/designsystem-css/_theme.scss'
import '@mattilsynet/designsystem-css/_color-tokens.scss'
import '@mattilsynet/designsystem-css/utilities/_mixins.scss'
import '@mattilsynet/designsystem-css/utilities/colors.js'
```
We can talk further about the filestructure and folders, but I think the project specific filestructure should not be exposed in the published package. What do you think?

TheBoosja (changes requested):

mt-kit/core/css/package.json
lha047: Yepp I agree :) Will try fix it :P

TheBoosja (approved):

PR #815, opened 2023-04-17
MTP-1472 lists a11y

lha047:


src/svelte/components/form/CheckboxWithSubSets.svelte
```diff
@@ -6,10 +6,13 @@
 
   export let params: Record<string, string> = {}
   export let name: string
+  let className = ''
+  export {className as class}
```
itemodahl, line 10: Is `class` used anywhere?

itemodahl (commented):
Good job.

src/svelte/components/form/CheckboxWithSubSets.svelte
lha047: Forgot to add it. Fixed now

itemodahl (approved):

PR #820, opened 2023-04-18
MTP-1491 Fix ariaLabel status

petterkoding:
- removed if statements to fix SR status update
- Changed class on span to class:spinner={inProgress}
- spinnerPlacement uses class with flexbox
- SR reads aria labels on WIN

The SR works fine on my end (except in IE), using a HP ZBook, WIN 11. Tested with NVDA, Jaws and Windows Narrator on Chrome, Edge, Firefox and IE(in Edge).


lha047 (approved):
Good job on refac and testing with SR :D

PR #833, opened 2023-04-27
MTP-546 Bytte ut knapper med lenker

petterkoding:
- Refactored link--transport class, new style modifiers: --primary, --secondary, -large.
- Updated CardArticle to use link--transport and removed unnecessary classes
- Removed unused styling classes on CardArticle.test.ts
- Updated Stories that used old styling to only use link--transport class


src/svelte/components/CardArticle.svelte
```diff
@@ -40,10 +40,7 @@
       <a
         href={linkUrl}
         rel={mapRelExternal(linkUrl)}
-        class:link--transport={linkTypeTransport}
-        class:color--primary={linkTypeTransport}
-        class:forward-arrow-end={linkTypeTransport}
-        class:no-underline={linkTypeTransport}>
+        class:link--transport={linkTypeTransport}>
```
petterkoding, line 43: is the CardArticle and its' anchor tag used anywhere without the **forward-arrow-end** icon at the end (just styled as default anchor tag)??

If not, then the exported prop `linkTypeTransport` isn't needed and can be removed to avoid forgetting to add this prop. We can just use `class="link--transport"` instead.

itemodahl (approved):
LGTM

src/svelte/components/CardArticle.svelte
lha047: I think you should remove the option. The cardArticle should always look the same

lha047 (changes requested):
Good Job :D 

Have a look at icons.
Could you clear up the link stories too?

src/scss/global/_links.scss
```diff
@@ -358,22 +358,104 @@ a {
...
+        --transportlink-tb-padding: 0.875rem;
+        @extend .no-underline;
+        color: var(--color-mt-brown);
+        transition: all ease-in-out 150ms;
+        padding: var(--transportlink-tb-padding) 0;
+        width: 100%;
+        border-radius: unset;
+        background-color: transparent;
+        border-top: var(--border-width-regular) solid var(--color-mt-beige);
+        border-bottom: var(--border-width-regular) solid var(--color-mt-beige);
```
petterkoding, line 447: Just realized this works for a single standalone link. If it's a list of links, then border-top and -bottom on each `<a>`  tag will create multiple lines between links. That's probably not what we want...

stories/content/ProfessionalAdvice.stories.svelte
```diff
@@ -61,13 +61,15 @@
           text="Noe tekst i kroppen"
           linkUrl="http://"
           linkText="Neste handling"
-          headerTag="h2" />
+          headerTag="h2"
+          linkTypeTransport={true}/>
```
lha047, line 65: Remove this?

stories/content/ProfessionalAdvice.stories.svelte
```diff
@@ -61,13 +61,15 @@
...
         <CardArticle
           type="task"
           title="Dette er en h3 tittel"
           text="Noe tekst i kroppen"
           linkUrl="http://"
-          linkText="Neste handling" />
+          linkText="Neste handling"
+          linkTypeTransport={true}/>
```
lha047, line 72: Remove this?

src/scss/global/_links.scss
```diff
@@ -435,19 +435,23 @@ a {
       border-radius: var(--border-radius);
 
       &-large{
+        --transportlink-tb-padding: 0.875rem;
         @extend .no-underline;
         color: var(--color-mt-brown);
-        transition: padding-left ease-in-out 150ms;
-        padding: var(--transportlink-padding);
+        transition: all ease-in-out 150ms;
```
lha047, line 441: Why do you need all here? isnt it just the padding-left that moves?

lha047 (changes requested):
Nice :D

Left some more comments

stories/content/Task.stories.svelte
```diff
@@ -126,6 +126,7 @@
         title={task.title}
         text={task.text}
         id={task.title}
+        linkTypeTransport={true}
```
lha047, line 129:
```suggestion
```

src/scss/global/_links.scss
```diff
@@ -358,22 +358,98 @@ a {
...
+      color: var(--color-primary);
+      border: var(--border-width-regular) solid var(--color-mt-beige);
+      font-weight: var(--font-weight-bold);
       padding: var(--transportlink-padding);
+      border-radius: $border-radius;
+
+      &-large{
+        @extend .no-underline;
+        color: var(--color-mt-brown);
+        transition: all ease-in-out 150ms;
```
lha047, line 439:
```suggestion
        transition: padding-left ease-in-out 150ms;
```

lha047 (approved):
Good job :D 

Fix comments before merging

PR #840, opened 2023-05-04
MTP-1554: upgrade checkboxwithsubsets

itemodahl:
MTP-1554

### DIFF
--- a/src/svelte/components/form/CheckboxWithSubSets.svelte
+++ b/src/svelte/components/form/CheckboxWithSubSets.svelte
@@ -9,6 +9,7 @@
   let className = ''
   export {className as class}
   export let legend: string
+  export let variation: 'primary' | 'secondary' = 'primary'
   export let options: Array<CheckboxWithSubSectionsOptions>
   export let categoryName = 'kategori'
   export let subCategoryName = 'underkategori'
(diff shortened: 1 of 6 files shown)


src/svelte/components/form/CheckboxWithSubSets.svelte
```diff
@@ -21,6 +22,9 @@
     }
   }
 
+  $: fieldsetClass =
+    variation === 'primary' ? 'checkbox-subsets--primary' : 'checkbox-subsets--secondary'
+
   interface CheckboxWithSubSectionsStates {
     key: string
     docCount: number
```
lha047, line 30: I think we need to make this optional

src/svelte/components/form/CheckboxWithSubSets.svelte
```diff
@@ -127,12 +130,12 @@
             on:change={() => mainCategory(mainIndex)} />
           <label for={listItem.key}>{`${listItem.displayName} (${listItem.docCount})`}</label>
```
lha047, line 131: Dont show the docCount when its not present. Same for the other docCount

lha047 (changes requested):
Nice job :D

See comments

src/svelte/components/form/CheckboxWithSubSets.svelte
itemodahl: What do you mean? It must be one of the two variations?

itemodahl: Ah you mean docCount...

PR #856, opened 2023-05-11
MTP-1554: Use absolute position to support safari

itemodahl:
MTP-1554


lha047 (changes requested):
Screenshot 2023-05-15 at 08 00 24
Doesnt seem to be working with new structure. And two lines of text

lha047:
@itemodahl should we close this?

PR #868, opened 2023-05-23
MTP-206 Remove Notice margins fix

petterkoding:
Notice box width should follow the rest of the content and use same width

Task: 
https://mattilsynet.atlassian.net/browse/MTP-206

### DIFF
--- a/src/scss/components/_notice.scss
+++ b/src/scss/components/_notice.scss
@@ -8,8 +8,6 @@
   background-color: #fff;
 
   @media only screen and (min-width: $breakpoint-large) {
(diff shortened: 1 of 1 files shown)


src/scss/components/_notice.scss
lha047, line 1: Remove the variable too if it is no longer in use

lha047: There is one more ref to `--notice-margin` on line 11

lha047 (changes requested):

lha047 (approved):
Nice :D

PR #874, opened 2023-06-01
MTP-1278 Top task fix

petterkoding:
- Removed `word-break: break-word` where its been used.

- Added `word-break: break-word` to elements where the text overflows in _typography.scss

related issue:
https://mattilsynet.atlassian.net/jira/software/projects/MTP/boards/30?assignee=63fc1ae07c30bbd6b33d9682&selectedIssue=MTP-1275


src/scss/global/_typography.scss
```diff
@@ -179,3 +179,11 @@ strong {
 .children-spaced-xxs {
   @include children-spaced(var(--spacer-xx-small));
 }
+
+h1,h2,h3,h4,h5,h6,
+table th,
+.summary-list,
```
lha047, line 185: Can we make it so that we can have `<a><button><ul><ol><dl><p>` in this list? and not the specific classes?

petterkoding: see latest commit, for the `<a>` and `<button>`  I've used `:not` to avoid breaking the language, search and menu words in the header.

src/scss/global/_typography.scss
```diff
@@ -179,3 +179,16 @@ strong {
 .children-spaced-xxs {
   @include children-spaced(var(--spacer-xx-small));
 }
+
+h1,h2,h3,h4,h5,h6,
+th,
+td,
+a:not(.button--link),
```
lha047, line 186: I would prefere not having exceptions on this. I think you can remove the `not()`-stuff if you change the `.button--link` a bit.
I tried to set button--link `min-width: max-content`. It seems to be ok. But on safari the header looks squashed. So I had to update `.header` in `_layout.scss` with `grid-template-columns: 1fr max-content max-content max-content;` then it seems to be ok.

lha047 (changes requested):
Looks nice :) 

Would prefere not to have exceptions on the word-beak rule. Have a look at my comment. Check if you can work with something like that

src/scss/global/_typography.scss
petterkoding: Agree, I also had that in mind after a while. Will take a look at your suggestions and go on from there

lha047 (approved):
Nice work :D

PR #875, opened 2023-06-02
MTP-1449 use z-index var on aside.expanded

petterkoding:
- Video covers aside menu

Note: 
**WIP** Go through all occurrences where z-index is used and replace with variables


lha047 (approved):
Nice :D

PR #897, opened 2023-07-05
MTP-1440 gjore underkapitler synlig i menyen veileder

lha047:


amundlindstol (changes requested):
LGTM, except that the icon does not follow the white selected state shown in figma.
Create new task, or fix here?
image

lha047:
The Figma sketch didnt account for the button being a separate button. I dont know how to style all the states for this light blue background. I think we leave it as is until someone complains

amundlindstol (approved):
👍🏻

PR #910, opened 2023-07-19
MTP-1681 Fix summarypage anchor breaks

petterkoding:
This fixes the issue where the anchor tag with text "Endre" breaks on the summary-step forms.


lha047 (approved):
Nice

PR #919, opened 2023-08-11
MTP-1664-ikon-for-horing

amundlindstol:


lha047 (approved):
Nice :D

mt-kit/core/css/src/components/_list.scss
```diff
@@ -157,3 +157,9 @@ ol {
     }
   }
 }
+
+.li-center-vertical {
```
lha047, line 161: Is this needed?

PR #922, opened 2023-08-21
MTP-1727: fix header layout safari

lha047:


Hannnst (approved):
Tested in safari, works great!

PR #923, opened 2023-08-24
MTP-1666 mtp 1657 svelte part for a vise liste av regelverksprosesser pa en temaside

petterkoding:
use with https://github.com/Mattilsynet/mattilsynet-web/pull/550

### DIFF
--- a/mt-kit/core/svelte/src/lib/svelte/components/Published.svelte
+++ b/mt-kit/core/svelte/src/lib/svelte/components/Published.svelte
@@ -6,7 +6,7 @@
   export let lang = 'NO-nb'
   export let publishedText = 'Publisert'
   export let lastPublishedText = 'Faglig oppdatert'
-  export let dtClass
+  export let dtClass = ''
   $: dateObject = compareDates({
     lang,
     publishFrom,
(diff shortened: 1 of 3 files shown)


mt-kit/core/svelte/src/lib/svelte/components/Published.svelte
```diff
@@ -6,6 +6,7 @@
   export let lang = 'NO-nb'
   export let publishedText = 'Publisert'
   export let lastPublishedText = 'Faglig oppdatert'
+  export let dtClass
```
lha047, line 9: Add default empty string so it doesn't put a undefined class in the string

petterkoding: ok

mt-kit/core/svelte/src/lib/svelte/components/Published.svelte
```diff
@@ -6,6 +6,9 @@
   export let lang = 'NO-nb'
   export let publishedText = 'Publisert'
   export let lastPublishedText = 'Faglig oppdatert'
+  export let dtIcon = ''
```
lha047, line 9: On components i would only use class for the root component.
To target other elements use eg. dtClass. Dont think you need two variables for putting different classes. Would only use that when there is theming sizing etc. and that is a part of the class name. I would not use the same name incase someone later dont want the same styling

lha047 (changes requested):
See comment

lha047 (approved):
Nice :D

PR #939, opened 2023-10-05
MTP-38 Fix Dialog boks paa veileder

petterkoding:


mt-kit/core/svelte/src/lib/svelte/components/DialogBox.svelte
```diff
@@ -14,6 +18,9 @@
 
   function handleClose() {
     isOpen = false
+    dispatch('closingDialog', {
```
lha047, line 21: Add type for the dispatched event
Place the type in the `types.d.ts`. Remember to reexport it to from the `index.ts` file to be able to use other places.

lha047 (changes requested):
Looks good :D 

Add typing for the event

amundlindstol (approved):
LGTM :D need me to test?

PR #948, opened 2023-10-26
MTP-1827 bruke riktige variabler i ds - lists

lha047:


TheBoosja (approved):
Is it on purpose that there are lines commented out like this: `<!-- <ul class="mt-ul">-->`? It seems only to be in `*.stories.svelte` files

lha047:
@TheBoosja no. Will fix

magnars (approved):

PR #963, opened 2023-11-08
MTP-1886 Fix footer scroll and flex direction on child elements

petterkoding:
fixes horizontal scroll caused by footer width and flex direction of child elements on breakpoint


lha047 (approved):
Nice :D 

I added some color ajustments

PR #974, opened 2023-11-16
MTP-1830 themes old new i ds - fix link color

lha047:


amundlindstol (approved):
Hard to see the white arrows in DS, but LGTM :D

PR #986, opened 2023-11-20
MTP-1830: fix update styling on heading

lha047:


mt-kit/core/svelte/src/lib/svelte/components/HeadingLevel.svelte
```diff
@@ -6,7 +6,9 @@
   let idPassed: string | undefined = undefined
   export { idPassed as id }
 
-  className = getClassName(className, headingLevel)
+  $: {
```
amundlindstol, line 9:
```suggestion
  $: className = getClassName(className, headingLevel)
```

amundlindstol (approved):
👍🏻 see suggestion

mt-kit/core/svelte/src/lib/svelte/components/HeadingLevel.svelte
lha047: didnt work. Think its because og the `export className as class` thing

PR #995, opened 2023-11-27
MTP-1862-utlisting-av-forsoksdyrsammendrag

amundlindstol:


petterkoding:
Tested in storybook. Looks good.

mt-kit/core/svelte/src/lib/svelte/components/Table.svelte
```diff
@@ -0,0 +1,28 @@
...
+  let className = ''
+  export { className as class }
+  export let style = ''
+  export let caption = ''
+  export let headers: Array<{}> = [{}]
+  export let rows: Array<{}> = [{}]
+</script>
+
+{#if caption}
+  <caption class="m-t-m">
```
lha047, line 11: I think we should pass spacing to component. Might not always need it

mt-kit/core/svelte/src/lib/svelte/components/Table.svelte
```diff
@@ -0,0 +1,28 @@
...
+    {caption}
+  </caption>
+{/if}
+<table class="mt-table responsive-table m-t-xxs {className}" id="table-{caption}" {style}>
```
lha047, line 15: `m-t-xxs` same as above

lha047 (changes requested):
Looks good :) 

Just a few comments

lha047 (approved):
Nice :D

PR #998, opened 2023-12-06
MTP-1951 sticky innholdsfortegnelse

petterkoding:
Caret down icon doesnt show because it's using the wrong color, hex instead of hsl. Looks like it doesnt exist in the color-tokens at the moment


amundlindstol:
The focus border should be more visible
image
Revert the changes you have made back to the original, but make focused white border the same size as the others.
image

amundlindstol:
The color should be dark when it's not open. Dont follow figma in this case
image

amundlindstol (changes requested):
changes needed

amundlindstol:
Menu should be on top of the chapter nav
image

mt-kit/core/css/src/components/_chapter-menu.scss
```diff
@@ -13,15 +13,20 @@
...
-      border-bottom: $border-width-small solid var(--color-mt-background);
+      border-bottom: none;
     }
     & + li.chapter-menu--chapter-wrapper {
       margin-top: 0;
     }
   }
+
+  h2{
+    display: none;
```
lha047, line 27: Why display none?

lha047 (changes requested):
Dont remove the heading. See comment

amundlindstol (approved):
LGTM :D

mt-kit/core/css/src/components/_chapter-menu.scss
petterkoding: "designer doesnt want to display "Innhold" twice." ...

My intentions was to hide, but not from a11y, just had to find and use correct class.

PR #1003, opened 2023-12-28
MTP-1677 Guide mobile chaptermenu with link, fix zindex bug + scroll …

petterkoding:
also fixes (prod bugs):

hides scrollbar gutter if theres no overflow, zindex issues with breadcrumbs and dialogue box overlapping mobile menu


mt-kit/core/css/src/components/_chapter-menu.scss
```diff
@@ -54,7 +58,7 @@
     color: var(--link-color);
     text-decoration: none;
     width: 100%;
-    padding: var(--spacer-xxx-small) 0;
+    padding: var(--spacer-xx-small) 0;
```
lha047, line 61: Why change this?

mt-kit/core/css/src/components/_chapter-menu.scss
```diff
@@ -67,6 +71,7 @@
     }
     &.chapter-menu--subchapter {
       padding-left: var(--spacer-xx-small);
+      border-bottom: none;
```
lha047, line 74: Dont think you need this. The border bottom is on the list item

mt-kit/core/css/src/components/_chapter-menu.scss
```diff
@@ -24,6 +24,10 @@
     }
   }
 
+  li:has(.chapter-menu--subchapter){
+    border-bottom: none;
```
lha047, line 28: Cool that you have looked into `:has()`. Just be aware that it doesnt work in older bowsers. In this case I it doesnt have that much of an impact. The older browsers will get the extra border bottom. Just be carefull when using new features and make sure it works for older browsers too

lha047:
Looks like the meny is sticky when you dont have JS. It has to behave as it used to when there is no JS. This implementation hides all the content as the menu always cover the entire page and you are not able to remove it

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -18,24 +18,51 @@ $mobile-menu-height: 3rem;
...
+  padding: var(--spacer-xx-small) var(--spacer-x-small);
+  background-color: var(--color-primary);
+
+  .chapter-title-link {
+    color: var(--color-mt-white);
+    height: 100%;
+    a {
+      padding-right: var(--spacer-x-small);
+    }
+    p {
```
lha047, line 38: Do you need this? I cant see where it is used

lha047: Is this for styling what is in [aria-expanded='false']?

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -18,24 +18,51 @@ $mobile-menu-height: 3rem;
...
+      line-height: theme.$font-size-small; color: var(--color-mt-white);
+    }
+  }
+
+  &[aria-expanded='true'] {
+    background-color: var(--chapter-menu-active-on-dark);
+    color: var(--color-primary);
+
+    p {
+      color: var(--color-primary);
```
lha047, line 49: Same as above. Cant see a p-tag inside the expanded

lha047:
Screenshot 2024-01-11 at 08 47 59
The link have to have focus styling

lha047:
Screenshot 2024-01-11 at 08 54 56
Can you fix the width and height of the close button on desktop? Should be the minimum width and height

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -18,24 +18,51 @@ $mobile-menu-height: 3rem;
...
-  position: sticky;
-  top: 0;
+
+  padding: var(--spacer-xx-small) var(--spacer-x-small);
+  background-color: var(--color-primary);
+
+  .chapter-title-link {
+    color: var(--color-mt-white);
+    height: 100%;
+    a {
```
lha047, line 35: Style on the class

mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -206,6 +206,19 @@ Only use if it does NOT affect accessibility
   }
 }
 
+/*
+Only use if it does NOT affect accessibility
+*/
+.responsive-show-flex-column-xx-large {
```
lha047, line 212: Dont think you need this. We should show the link on desktop as well to be consistent

lha047 (changes requested):
Good work
See comments. Make sure it works without JS as well.
Just ask if something is unclear

mt-kit/core/css/src/components/_chapter-menu.scss
petterkoding: This was to compensate for the "Veileder:" text above the title link. Looking at the figma design, it had more spacing above and under, but since its no longer applicable I´ll change back

mt-kit/core/css/src/components/_chapter-menu.scss
petterkoding: Was afraid of that, too bad it isn´t fully supported yet.

lha047 (changes requested):
Can you update the storybook Guide story so we can see how it works only in storybook

mt-kit/core/css/src/components/_chapter-menu.scss
```diff
@@ -34,8 +25,18 @@
     button.icon--caret-down-after-on-primary {
       grid-area: btn;
     }
-    ul {
+    .mt-ul {
       grid-area: sub;
+
+      .layout-flex {
```
lha047, line 31: I would add `.mt-li` to the list item and then use `.mt-li` for this styling. It doesnt seem right to use the `.layout-flex` for this kind of styling as it is a styling for the list item (and layout-flex can be for other things than list items)

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -112,12 +148,20 @@ $mobile-menu-height: 3rem;
   .layout-with-sidebar {
     overflow: hidden;
 
+    .aside-header {
```
lha047, line 151: Do you need this? The css is mobile first and should keep the display: none styling you added on line 19 until you use some other value for `display` in a media query for a larger size.

lha047 (changes requested):
Update the Guide story

lha047:
Screenshot 2024-01-22 at 08 24 53
Search link layout is broken

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -149,56 +184,47 @@ $mobile-menu-height: 3rem;
...
-        transition: width 0.3s ease, height 0.3s ease, top 0.3s ease;
+        top: 0;
+        left: 0;
+        position: static;
+        width: $width-sidebar-closed;
+        overflow: visible;
+        height: 100%;
+        transition: width 0.3s ease, height 0.3s ease;
+
+        //.mobile-menu {
```
lha047, line 200: REmove comment

mt-kit/core/css/src/components/_layout-with-sidebar.scss
```diff
@@ -10,70 +12,109 @@ $width-sidebar-closed: 3.5rem; //56px;
...
 
-  button {
-    padding-inline: var(--spacer-xx-small);
-    font-size: theme.$font-size-small;
-    line-height: theme.$font-size-small;
+  .mt-link {
+    @include focus-outline();
+    @include disabled;
+    color: var(--color-mt-white);
+    width: 75%;
```
lha047, line 47: I would create variables for the width here and the one with 25%. Place them underneath each other so it is easy to see what this is.

lha047 (changes requested):
Screenshot 2024-02-26 at 09 28 49
Can you fix the color of the x-button

PR #1026, opened 2024-01-29
MTP-2072: set focus only on interactions

lha047:
- to try to prevent interaction with map without focus on map on touch


amundlindstol (approved):
🤞🏻

PR #1047, opened 2024-02-15
MTP-1958 Newsletter checkbox issues

petterkoding:
Checkboxes getting squashed by long texts, MTP-1958

### DIFF
--- a/mt-kit/core/css/src/components/_radio-checkbox.scss
+++ b/mt-kit/core/css/src/components/_radio-checkbox.scss
@@ -25,6 +25,7 @@
   float: left;
   border-radius: 50%;
   color: var(--color-primary);
+  flex-shrink: 0;
 
   &:disabled,
   &[disabled] {
--- a/mt-kit/core/css/src/components/_radio-checkbox.scss
+++ b/mt-kit/core/css/src/components/_radio-checkbox.scss
@@ -25,7 +25,6 @@
   float: left;
   border-radius: 50%;
   color: var(--color-primary);
-  flex-shrink: 0;
 
   &:disabled,
   &[disabled] {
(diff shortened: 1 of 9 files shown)


mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -106,6 +106,7 @@
   display: flex;
   flex-direction: column;
   gap: var(--gap);
+  flex-wrap: nowrap;
```
lha047, line 109: This is going to change it for all the layout-flex-col. Did you check that it doesnt break any thing? Suggest you make a new class with this addition

mt-kit/core/css/src/components/_radio-checkbox.scss
```diff
@@ -25,6 +25,7 @@
   float: left;
   border-radius: 50%;
   color: var(--color-primary);
+  flex-shrink: 0;
```
lha047, line 28: Suggest you make a class for this in layout and use that. This class should not consider it self with layout based on parent

lha047 (changes requested):
See comments :)

mt-kit/core/css/src/components/_radio-checkbox.scss
petterkoding: appears that its no longer needed after using  `flex-wrap: nowrap` on the parent fieldset and on the nested fieldsets as well.

lha047 (changes requested):
Screenshot 2024-03-06 at 07 53 40
Screenshot 2024-03-06 at 07 53 30

Still breaks. Update stories to show how it breaks with long sentences in checkbox labels
Look into how the `.form-control` works with the grid. It should work with the correct values. You should not need the flex no-wrap. Check the html classes on checkboxwith subsets. Look into the displays in _checkboxes-with-subsets

amundlindstol (approved):
LGTM :D

PR #1061, opened 2024-03-04
MTP-1677 Guide title increase font size

petterkoding:
Designers wants bigger font to emphasize hierarchy of it being the title of the guide


lha047 (approved):
👍

PR #1064, opened 2024-03-15
MTP-2178 adjust header for mobile

petterkoding:
- add slightly larger search icon
- align items in header
- make logo smaller on mobile

logo is passed to Dropdown component in Web

see:  https://github.com/Mattilsynet/mattilsynet-web/pull/843


mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -260,6 +260,10 @@ Only use if it does NOT affect accessibility
   grid-template-columns: 1fr max-content max-content max-content;
   align-items: center;
 
+  >.mt-link {
```
lha047, line 263: Can you update the headers in the stories so we can see how this works?

lha047 (changes requested):
Have a look at comments. 

Could you also fix the no-js styling?
Screenshot 2024-03-20 at 07 53 39

lha047 (changes requested):
Good work :)

Did you look at the styling of the header without JS?

petterkoding:
> Good work :)
> 
> Did you look at the styling of the header without JS?

See the other PR for styling without JS

lha047 (approved):
👍

PR #1066, opened 2024-03-18
MTP-2139 adjustments on search

petterkoding:


lha047 (changes requested):
Look into if you can remove/change the already existing code for search.

lha047 (changes requested):
Screenshot 2024-03-21 at 08 46 53
Screenshot 2024-03-21 at 08 47 03
Looks like there is too much border on the hover effect

PR #1067, opened 2024-03-19
MTP-2179 Add shadow when menu is open

petterkoding:


mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -422,3 +422,16 @@
     }
   }
 }
+.mt-header:has(.visible) + .container-full-width {
```
lha047, line 425: Screenshot 2024-03-20 at 09 52 50
Cant have this styling if there is no-js

lha047 (changes requested):
See comment. 

Update header story to show how this works

mt-kit/core/svelte/src/lib/svelte/components/Dropdown.svelte
```diff
@@ -35,7 +35,10 @@
   }
 </script>
 
-<div class="dropdown {className} {onServer ? 'on-server' : ''}" class:visible={isOpen || onServer}>
+<div
```
lha047, line 38: You dont need to change this. It is possible to do this with the .visible class and the `details[open]` and the [aria-expanded='true'] on the button when js is on.

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -422,3 +426,18 @@
     }
   }
 }
+.mt-header:has(.visible) + .container-full-width,
```
lha047, line 429: You can use the [aria-expanded="true"] on the button to get the same effect as details[open]

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -194,7 +198,7 @@
     }
   }
 
-  &.visible {
+  &.visible, &.visible-no-js {
```
lha047, line 201: Dont need .visible-no-js. See other comment

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -184,7 +188,7 @@
       .disclosure-header {
         font-size: $font-size-h3;
       }
-      &.visible {
+      &.visible, &.visible-no-js {
```
lha047, line 191: Dont need .visible-no-js. See other comment

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -164,6 +167,7 @@
       }
     }
     &.visible,
+    &.visible-no-js,
```
lha047, line 170: Dont need .visible-no-js. See other comment

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -90,6 +90,7 @@
     position: relative;
     z-index: 0;
     margin-bottom: $border-width-large;
+    display: inline-flex;
```
lha047, line 93: You dont need this. The .mt-button is already inline-flex

lha047 (changes requested):
You dont need to refac add the `.visible-no-js` to this to achieve what we are looking for.  See comments

mt-kit/core/css/src/components/_dropdown.scss
lha047: You only need the display: inline-flext on the .mt-summary.mt-button

lha047 (changes requested):

lha047 (approved):
Good job :D

PR #1068, opened 2024-03-21
MTP-2171 new pagination component #minor

lha047:


mt-kit/core/svelte/src/lib/svelte/components/pagination/Pagination.svelte
```diff
@@ -0,0 +1,174 @@
...
+  export let currentPageIndex: 0 | 1 = 0
+
+  let className = ''
+  export { className as class }
+
+  $: nextPageIndex = currentPageIndex + 1
+  $: nextPage = pages[nextPageIndex]
+  $: previousPageIndex = currentPageIndex - 1
+  $: previousPage = pages[previousPageIndex]
+  $: media = undefined
```
amundlindstol, line 23: seems like you only use media for media?.matches.
Could you create a bool var with a better name?

amundlindstol (changes requested):
When navigating on desktop, the numbers "jump" when going from 2 to 3
image
image
I would prefer if the current page only moves one step at the time.

Also, there's room to have every number 1 through 7 displayed without any truncate.

This should not happen, right?
image

mt-kit/core/svelte/src/lib/svelte/components/pagination/Pagination.svelte
```diff
@@ -0,0 +1,157 @@
...
+          <li
+            class="mt-li pagination-item"
+            class:pagination-item--current={index === currentPageIndex ? 'page' : undefined}>
+            <a
+              href={chapter.url}
+              class="mt-link"
+              title={interpolate(toPageTitle, [index + 1])}
+              aria-current={index === currentPageIndex ? 'page' : undefined}
+              on:click|preventDefault={() => handleClick(index)}>
+              <span class="inclusively-hidden-initial">{labelPage}</span>
```
petterkoding, line 127: Screenreaders wont notify if its the last page which may disorient users. I suggest changing to "page x of y" as discussed

petterkoding (changes requested):
nice work! Works smoothly. Take a look at label for screenreaders

mt-kit/core/svelte/src/lib/svelte/components/pagination/Pagination.svelte
lha047: Fixed

amundlindstol (approved):
LGTM :D

PR #1070, opened 2024-03-21
MTP-2183 focus search when dropdown opens

petterkoding:
pass open state to slot and call focus if true


mt-kit/core/svelte/src/lib/svelte/components/form/Search.svelte
```diff
@@ -17,6 +18,10 @@
   let className = ''
   export { className as class }
 
+  let ref: HTMLInputElement
```
lha047, line 21: Maybe rename this to searchInput or something? for more descriptive naming

lha047 (approved):
Nice work :D Consider comment before merging

PR #1082, opened 2024-04-09
MTP-2204 close dropdown when focus leaves

petterkoding:


mt-kit/core/svelte/src/lib/svelte/components/Dropdown.svelte
```diff
@@ -33,6 +33,20 @@
   if (loadJs) {
     onMount(() => send('MOUNTED'))
   }
+  function handleTabOut(node: HTMLElement, callBack: () => void): { destroy(): void } {
```
lha047, line 36: Extract this function as a separate action. You can create a new file in the ts folder. Look at how its done with the `clickOutside-action`. It should also be exposed in from the ts folder so it can be used separately

lha047 (changes requested):
Nothing hapens when I tab from the last link in the menu. Maybe update the story?
See comment about moving the action function

mt-kit/core/svelte/src/lib/ts/focus-outside.ts
lha047, line 1: Can you add this to the exports in /src/lib/ts/index.ts

lha047 (changes requested):

lha047 (approved):
👍

PR #1084, opened 2024-04-15
Mtp 2211 designjusteringer

petterkoding:
All page numbers gets circle on hover
Increase padding on next/previous buttons + border radius
Increase spacing between pagination and last result
Change hover background color on results

### DIFF
--- a/mt-kit/core/css/src/components/_pagination.scss
+++ b/mt-kit/core/css/src/components/_pagination.scss
@@ -64,7 +64,6 @@
         background-color: var(--mt-button-primary-bg-hover);
         color: var(--mt-button-primary-color-hover);
         border-color: var(--mt-button-primary-border-hover);
-        border-radius: var(--border-radius);
       }
     }
     &.next-link {
--- a/mt-kit/core/css/src/components/_pagination.scss
+++ b/mt-kit/core/css/src/components/_pagination.scss
@@ -47,9 +47,15 @@
     display: flex;
     justify-content: center;
     align-items: center;
+    &:hover {
+      background-color: var(--mt-button-primary-bg-hover);
+      color: var(--mt-button-primary-color-hover);
+    }
+
     &.next-link,
     &.previous-link {
-      border-radius: 0;
+      padding: 0 var(--spacer-xxx-small);
+      border-radius: var(--border-radius);
       background-color: transparent;
       border-color: var(--mt-button-primary-border);
       text-decoration: none;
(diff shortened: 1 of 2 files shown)


mt-kit/core/css/src/components/_pagination.scss
```diff
@@ -58,6 +64,7 @@
         background-color: var(--mt-button-primary-bg-hover);
         color: var(--mt-button-primary-color-hover);
         border-color: var(--mt-button-primary-border-hover);
+        border-radius: var(--border-radius);
```
lha047, line 67: Do you need this? Its already set on default

mt-kit/core/css/src/global/_links.scss
```diff
@@ -303,7 +303,7 @@ $transportlink-colors: 'error' !default;
 
     text-decoration-line: none;
     &:hover {
-      background-color: var(--mt-button-primary-bg-hover);
+      background-color: var(--colored-card-bg);
```
lha047, line 306: I would use eg. `--mt-button-flat-bg-hover` if you want `--color-mt-bg-furu` here. As the varaiable name is closer to what you want.

mt-kit/core/css/src/components/_pagination.scss
```diff
@@ -47,9 +47,15 @@
     display: flex;
     justify-content: center;
     align-items: center;
+    &:hover {
+      background-color: var(--mt-button-primary-bg-hover);
```
lha047, line 51: Looks like the hover-effect on pagination elements are --mt-button-flat-bg-hover (bg-furu) not the primary bg hover. Check all elements

lha047 (changes requested):
See comments

mt-kit/core/css/src/components/_pagination.scss
petterkoding: i dont, just forgot to delete it

mt-kit/core/css/src/global/_links.scss
petterkoding: fixed

mt-kit/core/css/src/components/_pagination.scss
petterkoding: even on the current page? Might have to ask designer about it. Since its about consistency on hover background-color.

petterkoding: I'll follow the figma designs as they use --mt-button-flat-bg-hover for bg on hover state for all elements

lha047 (approved):
Nice :D

PR #1088, opened 2024-05-03
MTP-2211 design adjustments padding and spacing inside mt-form elements

petterkoding:


mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -184,8 +184,22 @@
...
+      }
+
+      .mt-button {
+        margin-top: var(--spacer-x-small);
+      }
+    }
+
+    &.visible, .mt-details[open] {
+      >.mt-button {
+        box-shadow: 0 0 0 0.125rem var(--color-mt-background), 0 0 0 0.35rem var(--mt-button-primary-bg-focus);
```
lha047, line 201: Should there be a variable for this? Or do we have something that can be used?

mt-kit/core/css/src/components/_dropdown.scss
```diff
@@ -184,8 +184,22 @@
...
+      }
+
+      .mt-button {
+        margin-top: var(--spacer-x-small);
+      }
+    }
+
+    &.visible, .mt-details[open] {
+      >.mt-button {
+        box-shadow: $focus-box-shadow;
```
lha047, line 201: I dont think this should have the focus marking. (Same goes for the JS version)
Focus is for focus and this is a different state. The arrow on the dropdown indicates that the dropdown is open. Its confusing that there is focus on the whole dropdown and inside the filter (or what ever is inside the dropdown)

petterkoding: Id agree with you that it feels off and breaks away from how the rest of the content behaves on focus. Like the focus ring is stuck while it also continues to the next element inside.
The arrow and the content that slides out or appears (noJs) should be a clear indicator that its open/toggled.

Ive tried with just increasing the border thickness, but it still feels off.

image

(increased border from small to regular)

Should this be removed then?

lha047: I think we should not use any sort of outline/focus. In some cases the focus is just a border like this so i think its too much alike the focus indicator

lha047 (approved):
Nice 👍

PR #1089, opened 2024-05-22
MTP-2241 Adjust form spacing

petterkoding:
adjust spacing (padding) around help text and above expandable input list


lha047 (approved):
NICE 👍

PR #1092, opened 2024-05-23
MTP-2263 Fix invisible skiplink, Create skip-link class, adjust z-index on mt-header

petterkoding:


mt-kit/core/css/src/global/_body.scss
```diff
@@ -155,6 +155,7 @@
   --border-width-small: #{$border-width-small};
   --border-width-regular: #{$border-width-regular};
   --z-index-dropdown: #{$zindex-dropdown};
+  --z-index-fixed: #{$zindex-fixed};
   --z-index-feedback: #{$zindex-tooltip};
```
lha047, line 159: Could you clean up this? These should not be assigned to something with a different name. Makes it confusing.

lha047 (changes requested):
The skip-link is an accessibility feature that should always be available. I suggest making a new variable with apropriate naming and use that for the skip link. Make sure to test with the menu in guide as well

lha047 (approved):

PR #1100, opened 2024-06-17
MTP-2332 Update formsteps to handle "hidden" pages

lha047:
- logikk for a sette opp ny side fra betingelse


amundlindstol (approved):
LGTM :D

PR #1109, opened 2024-07-09
MTP-2407 fremhevet innhold endringer

petterkoding:
added $transition-ms-... variables to use for smoother background-color transitions

### DIFF
--- a/mt-kit/core/css/src/global/_links.scss
+++ b/mt-kit/core/css/src/global/_links.scss
@@ -367,8 +367,8 @@ $transportlink-colors: 'error' !default;
         }
       }
     }
-    &--blue {
-      color: var(--highlighted-blue-color);
+    &--cta {
+      color: var(--color-mt-text-dark);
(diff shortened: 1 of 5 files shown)


mt-kit/core/css/src/global/_links.scss
```diff
@@ -423,6 +426,10 @@ $transportlink-colors: 'error' !default;
       }
     }
 
+    &.b-r-radius {
```
lha047, line 429: There is already a class that does this. `.box-border-radius-br-l`

lha047 (changes requested):
You need to update all variations of fremhevet innhold. 
Should it have the big border radius? See comment

mt-kit/core/css/src/global/_links.scss
petterkoding: will change and delete the class i created

lha047 (approved):
Good work :D

PR #1113, opened 2024-07-15
MTP-2395 transport link endringer

petterkoding:
transport link styling
default
primary
primary--large
secondary
secondary--large

+

used as link list:
Lighter border color


mt-kit/core/css/src/global/_links.scss
```diff
@@ -428,104 +428,112 @@ $transportlink-colors: 'error' !default;
...
 
     &:focus {
       border-color: var(--transportlink-background);
       box-shadow: $focus-box-shadow;
     }
-    &:hover {
-      background-color: transparent;
-    }
 
     &--primary {
```
lha047, line 454: For reuse I would create variables for default, hover and active and use in the &--primary and &-large. Same for the states in secondary. Maybe include variation and state in the names

mt-kit/core/css/src/global/_links.scss
```diff
@@ -428,104 +428,112 @@ $transportlink-colors: 'error' !default;
...
+      }
+      &:active {
+        --transportlink-background: var(--color-mt-lys-furu);
+        @media (prefers-reduced-motion: no-preference) {
+          transition: background-color $transition-ms-300 ease;
+        }
+      }
 
       &-large {
-        @extend .no-underline;
```
lha047, line 516: Missing border radius
Screenshot 2024-08-01 at 12 11 40

mt-kit/core/css/src/global/_links.scss
```diff
@@ -428,104 +428,112 @@ $transportlink-colors: 'error' !default;
     }
   }
   &.link--transport {
-    --transportlink-background: var(--color-#{$transportlink-default-color});
```
lha047, line 431: Missing underline and state styling
Screenshot 2024-08-01 at 12 11 48

lha047 (changes requested):
See comments

mt-kit/core/css/src/global/_links.scss
```diff
@@ -428,104 +428,112 @@ $transportlink-colors: 'error' !default;
...
+      --transportlink-background: var(--color-mt-green-300);
       padding: var(--transportlink-padding);
-      border-radius: $border-radius;
-
-      &:after {
-        background-image: var(--icon-arrow-primary);
-      }
 
       &:hover {
-        color: var(--mt-button-primary-color-hover);
```
lha047, line 474: Why not use the mixins?

mt-kit/core/css/src/global/_links.scss
petterkoding: the default and `--primary` and `--secondary` (not large versions) will inherit text-decoration from the `.mt-link` class. Adjusting it thereafter by removing on hover (default) and setting to `none` on --primary and --secondary.

lha047 (changes requested):
Screenshot 2024-08-05 at 13 53 27
Looks like you need som gaps after you removed margin on the psudo

mt-kit/core/css/src/utilities/colors.js
```diff
@@ -135,7 +135,7 @@ module.exports = [
   },
   {
     name: 'mt-lys-furu',
-    color: '#8AC2AE',
+    color: '#9BCAB9',
```
petterkoding, line 138: make global vars for new link colors

lha047 (approved):
👍

PR #1116, opened 2024-07-16
MTP-2406 lenkeliste

petterkoding:


mt-kit/core/css/src/global/_links.scss
```diff
@@ -138,20 +139,13 @@ $transportlink-colors: 'error' !default;
   }
 
   &.forward-arrow-end-link {
```
lha047, line 141: If we're not having any links with arrows after. Maybe remove the classes with "end" and update?

petterkoding: Figma has these examples though

lha047: ok

mt-kit/core/svelte/src/stories/Links.stories.svelte
```diff
@@ -223,7 +223,7 @@
     <ul class="mt-ul m-t-xxs layout-grid list-unstyled">
       {#each cards as link}
         <li>
-          <Link href={link.href} class="forward-arrow-end-link" linkText={link.text} />
+          <Link href={link.href} class="forward-arrow" linkText={link.text} />
```
lha047, line 226: Screenshot 2024-08-02 at 09 46 56
Shouldnt the hover effect be the slightly larger underline?

lha047 (changes requested):
See comment

lha047 (approved):
👍 looks like it will be ok when the general link styling is applied

mt-kit/core/svelte/src/stories/Links.stories.svelte
petterkoding: MTP-2397 will fix it

PR #1117, opened 2024-07-17
MTP-2437 change link color Footer and add more space, update story

petterkoding:


mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -302,17 +302,22 @@ Only use if it does NOT affect accessibility
     display: flex;
     flex-direction: column;
     justify-self: center;
-    gap: var(--spacer-x-small);
+    gap: var(--spacer-medium);
     a {
-      color: var(--mt-footer-links-color);
+      color: var(--mt-footer-color);
```
lha047, line 307: This has the same challenge with color as the links in the text. Lets ask about this too. Also use the --mt-footer-links-color variable if it needs to change color

lha047 (changes requested):
See comment

lha047 (approved):
LGTM 👍

PR #1118, opened 2024-07-19
MTP-2392 remove padding-left on hover

petterkoding:
Remove all places padding-left is used on hover links and Disclosures etc.

### DIFF
--- a/mt-kit/core/css/src/global/_links.scss
+++ b/mt-kit/core/css/src/global/_links.scss
@@ -420,11 +420,7 @@ a {
     }
   }
   &.animated-header-arrow-after {
-    transition: padding-left ease-in-out 150ms;
     width: 100%;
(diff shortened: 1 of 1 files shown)


mt-kit/core/css/src/global/_links.scss
```diff
@@ -432,22 +419,6 @@ a {
       }
     }
   }
-  &.animated-header-arrow-after {
-    transition: padding-left ease-in-out 150ms;
-    width: 100%;
```
lha047, line 437: This width is maybe important to keep. Did you check that this will be ok when this class is removed?

petterkoding: I cant find it in use, web and ds. Have you?

lha047 (approved):
👍

PR #1119, opened 2024-07-19
Mixin for applying icons

petterkoding:
Making it easier to apply icons to links and buttons + disclosure-header

@include <mixin-name>($icon, $position, $dir, $spacer);

$icon: icon to use
$position: before or after
$direction (optional): it's possible to rotate the icon based on input "down, up, left and right", no need to add 4 arrows for 4 directions.
$spacer: space between icon and text

- nested @content in ::before/::after block makes it possible to write additional styles for the pseudo element
- you can add two icons on a link by using the mixin twice on a class with different $dir values
- hover effects such as spinning is easy to do with css on the class itself, also works for the [aria-expanded='true'] by calculating the $direction and subtracting 180 degrees to make it 'spin' in the opposite direction


mt-kit/core/css/src/components/_disclosure.scss
```diff
@@ -81,68 +88,29 @@
     }
 
     .disclosure-header {
-      transition: padding-left ease-in-out 150ms;
-
-      &:hover {
-        padding-left: var(--disclosure-padding-hover);
+      @include apply-icon($icon-carret-down-primary, before) {
```
lha047, line 91: include apply-icon in the main disclosure-header selector

lha047 (changes requested):
See comments

lha047 (approved):
Good job with this mixin :D

PR #1122, opened 2024-08-02
MTP-2444 fix width and icon alignment, update mixin, story 

petterkoding:


mt-kit/core/css/src/utilities/_mixins.scss
```diff
@@ -236,6 +236,7 @@
     height: $icon-size-x-large;
     margin-#{$margin-side}: $spacer;
     transform: rotate(#{$dir-mapped-to-rotate}deg);
+    flex-shrink: 0;
```
lha047, line 239: You dont know that there is a flexbox parent to this. Does not belong in the mixin. If you need it somewhere pass it as content

petterkoding: moved it. Also added fix for disclosures causing icon width to be smaller and off center

mt-kit/core/css/src/components/_disclosure.scss
```diff
@@ -17,6 +17,7 @@
   .disclosure-header {
     @include apply-icon($icon: $icon-carret-down-primary, $position: before) {
       margin-left: var(--disclosure-spacing);
```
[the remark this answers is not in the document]
petterkoding: agreed, will remove margin from mixin and use gap on main element.

lha047 (approved):
LGTM 👍

PR #1123, opened 2024-08-02
MTP-2442 add rule for disclosures used as link, move mt-summary logic

petterkoding:
:not([href]) check before using mixing on .disclosure-header, to exclude if it is a link

Add extra class to not cause issue for other places mt-summary is used

Mixin is moved from:
.mt-summary -> &.mt-summary-icon

### DIFF
--- a/mt-kit/core/svelte/src/lib/svelte/components/SummaryDetail.svelte
+++ b/mt-kit/core/svelte/src/lib/svelte/components/SummaryDetail.svelte
@@ -22,7 +22,7 @@
   aria-labelledby={ariaLabelledBy}
   data-test-id={testId}
   bind:open={isOpen}>
-  <summary class="mt-summary {summaryClass}">{@html title}</summary>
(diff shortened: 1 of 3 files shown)


lha047 (changes requested):
Screenshot 2024-08-05 at 14 06 42
Remove down caret on the menu items that are not expandable

mt-kit/core/svelte/src/lib/svelte/components/SummaryDetail.svelte
```diff
@@ -22,7 +22,7 @@
   aria-labelledby={ariaLabelledBy}
   data-test-id={testId}
   bind:open={isOpen}>
-  <summary class="mt-summary {summaryClass}">{@html title}</summary>
+  <summary class="mt-summary mt-summary-icon {summaryClass}">{@html title}</summary>
```
petterkoding, line 25: add class to macro

lha047 (approved):
👍 LGTM

PR #1127, opened 2024-08-13
MTP-2459 Change link icons in Menu, update Story

petterkoding:
Disclosures used in Header changes:

- icon should be on right side
- if disclosure is used as a link use arrow icon
- default text link styling for child links and bottom row links (underline with 3px on hover)
- add hover background-color and for open states
- icon background should change to white on hover/open state

image


lha047 (changes requested):
Screenshot 2024-08-14 at 08 41 21
Talk to Emilie about how they want it

mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -265,6 +265,42 @@ Only use if it does NOT affect accessibility
   border-bottom: $border-width-regular solid var(--mt-header-footer-border);
   z-index: var(--z-index-fixed);
   position: relative;
+
+  .dropdown-content {
```
lha047, line 269: Dont think this is the right place to do this

lha047:
Screenshot 2024-08-15 at 14 09 24
The hover effect looks wrong. Should find another solution

petterkoding:
> Screenshot 2024-08-15 at 14 09 24 The hover effect looks wrong. Should find another solution

Do you mean because of the icon has white background on hover/open state?

lha047 (approved):
LGTM 👍

PR #1132, opened 2024-08-14
MTP-2438 Fix for disclosure in layout-with-sidebar

petterkoding:
fixes the issue where ::before element is treated as grid item

image


lha047 (approved):
👍

PR #1133, opened 2024-08-14
MTP-2462 editor controlled hideKeywords, add if check, types

petterkoding:


petterkoding:
closing this PR because we have found another solution to just remove the fallback when setting returning keywords

PR #1134, opened 2024-08-15
MTP-2461 text-decoration not use shorthand to support Safari

petterkoding:


mt-kit/core/css/src/global/_links.scss
```diff
@@ -21,8 +21,11 @@ a {
   padding: var(--border-width-regular);
   line-height: 1.75rem;
   min-width: $minimum-click-width;
-  text-decoration: $color-link-underline underline $border-width-regular;
+  text-decoration-line: underline;
+  text-decoration-color: $color-link-underline;
+  text-decoration-thickness: $border-width-regular;
   text-underline-offset: $spacer-xxx-small;
+  text-decoration-skip-ink: auto;
```
lha047, line 28: Shouldnt this be `none`?

petterkoding: ah, thats correct! Correcting now

lha047 (approved):
👍

PR #1139, opened 2024-08-29
MTP-2493: update checkbox

lha047:
- to avoid rerender when value is undefined
- dispatching custom event for previous onChange event


amundlindstol (approved):
Seems to be working for me :D

PR #1142, opened 2024-09-06
MTP-2519 fix external links in disclosure breaks

petterkoding:


petterkoding:
moving fix to MTP-2491 branch

PR #1151, opened 2024-09-16
MTP-2548 apply external icon on Highlighted content and -links

petterkoding:


mt-kit/core/css/src/global/_links.scss
```diff
@@ -427,6 +428,11 @@ a {
     @include apply-icon($icon-arrow-right-primary-small, before) {
       flex-shrink: 0;
```
lha047, line 429: Did you check if you needed this? Looks to me like its not needed

mt-kit/core/css/src/global/_links.scss
```diff
@@ -427,6 +428,11 @@ a {
...
     }
+    @include apply-external-link-icon($position: before) {
+      &::after {
```
lha047, line 432: Did you check if you needed this? Looks to me like its not needed

lha047 (changes requested):
See comments

mt-kit/core/css/src/global/_links.scss
petterkoding: it is needed for mobile

image

mt-kit/core/css/src/global/_links.scss
petterkoding: actually theres a  <p> wrapping the <a> in CardArticle.svelte. I can just remove the <p> Then there wont be a second icon.

lha047 (changes requested):

PR #1155, opened 2024-09-19
MTP-2553 refac apply-icon mixin, use external icon logic internally

petterkoding:


mt-kit/core/css/src/utilities/_mixins.scss
```diff
@@ -269,36 +246,98 @@
...
+  $icon-direction,
+  $mapped-rotation
+) {
+  @if $position == before {
+    &:after,
+    &::after {
+      display: none;
+    }
+  }
+  @if $position == after {
```
lha047, line 322: Isnt there only two values for $position? Can you use if else?

mt-kit/core/css/src/utilities/_layout.scss
```diff
@@ -305,7 +305,7 @@ Only use if it does NOT affect accessibility
     justify-self: center;
     gap: var(--spacer-medium);
     a {
-      @include apply-external-link-icon($icon: $icon-external-link-on-dark, $position: after);
+      @include apply-icon($position: after, $overwrite-external-icon: true);
```
lha047, line 308: Like that you provide option to overwrite :D 
Would it be better to pass the desired icon to the $overwrite-externa-icon? It might not be you just want one option when overwriting

mt-kit/core/css/src/utilities/_mixins.scss
```diff
@@ -269,36 +246,97 @@
...
-      display: inline-flex;
-      align-items: flex-start;
-      gap: $spacer-xxx-small;
+      @include apply-pseudo-styling(
+        $icon: $icon,
+        $position: $position,
+        $icon-direction: $direction,
+        $mapped-rotation: $dir-mapped-to-rotate
+      );
+      @content;
```
lha047, line 295: Wont this content be applied to the html element (not the pseudo)? If you want to target the element just add it to the css classes no need to pass it here. Shouldnt the content be applied to the pseudo?

mt-kit/core/css/src/utilities/_mixins.scss
```diff
@@ -269,36 +246,97 @@
...
+    width: $icon-size-x-large;
+    height: $icon-size-x-large;
+    vertical-align: middle;
+    @if ($icon-direction != 'down') {
+      transform: rotate(#{$mapped-rotation}deg);
+    }
+    @media (prefers-reduced-motion: no-preference) {
+      transition: transform $transition-ms-300 ease;
     }
-    @content;
```
lha047, line 302: Isnt this what you are looking to do? See comment above

mt-kit/core/css/src/global/_links.scss
```diff
@@ -8,10 +8,15 @@ $transportlink-default-color: 'primary' !default;
 $transportlink-colors: 'error' !default;
 
 a {
-  @include apply-external-link-icon($position: before);
+  &:not(p &) {
```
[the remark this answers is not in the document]
petterkoding: by excluding theres no need to hide the other "side" of the pseudo element with display:none; It will also use the mixin twice on  a + (p > a).

But then theres occasions where we need to hide the opposite pseudo to fix design problems.

mt-kit/core/css/src/global/_links.scss
```diff
@@ -8,10 +8,15 @@ $transportlink-default-color: 'primary' !default;
...
+    @include apply-icon($position: before);
+  }
+}
 
-  p & {
-    @include apply-external-link-icon($position: after);
+.text,
```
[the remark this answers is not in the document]
petterkoding: theres articles on lovdata that doesnt use <p> tags, just free text with links, I have added the .text class to those on lovdata-wrapper to handle external links on appeal-decision, legal-guidance and collection-view

mt-kit/core/css/src/utilities/_mixins.scss
petterkoding: I think i forgot I refactored the original mixin and not passing it to the actual pseudo :)

mt-kit/core/css/src/global/_links.scss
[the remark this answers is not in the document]
lha047: The lovdata stuff goes in the web project

mt-kit/core/css/src/global/_links.scss
```diff
@@ -221,7 +225,7 @@ a {
     }
   }
   &.top-task {
-    @include apply-external-link-icon($position: after, $target-child: 'h2, h3');
+    @include apply-icon($position: after, $target-child: 'h2, h3');
```
lha047, line 228: The icon looks a bit low

lha047 (approved):
Good job 👍

PR #1158, opened 2024-09-25
MTP-2506 breadcrumbs fix and searchresult crumbs align

petterkoding:
ellipsis is now working for the last crumb, but its a problem with the second last, cause of the wrapper styles causes the text/icon to be shrinked down and hidden when scaling down the window


lha047 (approved):
👍

PR #1161, opened 2024-10-01
MTP-2570 lenkene pa hoyresiden har feil font str

petterkoding:
Links on right side in Menu has wrong font-size (28px), should be 24px on desktop and 21 on mobile. Bottom and right side should always be the same.
split up styles between disclosures and normal links which is not disclosures and doesnt have icon svgs


lha047 (approved):
Good job 👍

PR #1172, opened 2024-11-01
MTP-2651: remove special styling on card-article intro

lha047:
- content designers say we should not use italic for this.


amundlindstol (approved):
👍🏻

PR #1179, opened 2024-11-26
MTP-2780 upgrade sass ds

petterkoding:


amundlindstol (approved):
👍🏻

lha047 (changes requested):
Can we fix all the warnings in the console? If we dont do it now we only create more work later

PR #1180, opened 2024-11-26
MTP-2769 fix console warnings after upgrading sass

petterkoding:
deprecation warning about Legacy JS api is fixed in `svelte-preprocess 6.0.3`

explained here
https://github.com/sveltejs/svelte-preprocess/blob/main/CHANGELOG.md

### DIFF
--- a/mt-kit/core/css/src/components/_breadcrumbs.scss
+++ b/mt-kit/core/css/src/components/_breadcrumbs.scss
@@ -42,7 +42,9 @@
       }
       .mt-button--link {
         @include text-decoration-underline;
-        border-radius: 0;
+        & {
+          border-radius: 0;
+        }
         &:focus {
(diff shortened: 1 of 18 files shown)


lha047 (changes requested):
This needs to be in the same PR as the upgrade. I cant look at this when the dependency is not here

mt-kit/core/css/src/components/_breadcrumbs.scss
```diff
@@ -42,18 +42,24 @@
       }
       .mt-button--link {
         @include text-decoration-underline;
-        border-radius: 0;
+        & {
```
lha047, line 45: Why do we need this? Seems unnecesary when it compiles down to the same as it would have without the `& { ... }`

PR #1183, opened 2024-11-29
MTP-2624: add name to option id and label for

lha047:
- to make the options unique and avoid selecting option on other input with same options name


amundlindstol (approved):
LGTM :D

PR #1184, opened 2024-11-29
MTP-2785: close summary details when summaryId changes

amundlindstol:


lha047 (approved):
👍 
Seems strange that we have to do this

PR #1200, opened 2024-12-11
MTP-2629: rename slot headers to header #minor

lha047:
- cant have same name as component prop in svelte 5


amundlindstol:
I named it headersSlot and rowSlot in 2803 to make it stand out. what do you think?

PR #1201, opened 2024-12-13
MTP-2799 remove xstate from dropdown

petterkoding:


petterkoding:
@amundlindstol nice, works as expected!

amundlindstol (approved):
👍🏻

PR #1207, opened 2025-01-07
MTP-2803 Revert use of required on radio as it prevents submit

kajsaeggum:
Too ambitious when cleaning warnings, this topic will have to be addressed when we have uu-hat on later instead


lha047 (approved):
👍

PR #1208, opened 2025-01-09
MTP-2803 Refactored loadJs check in CheckboxWithSubSets

kajsaeggum:
Also fixed error with 3rd level options children: when no js is loaded they weren't shown until click, but should be permanently shown just like level 1 and 2

### DIFF
--- a/mt-kit/core/svelte/src/stories/form/CheckboxWithSubSets.stories.svelte
+++ b/mt-kit/core/svelte/src/stories/form/CheckboxWithSubSets.stories.svelte
@@ -5,81 +5,83 @@
   import { wrapInShadowDom } from '../storybook-utils/utils'
   import { interpolate, toKebabCase } from '$lib/ts/utils'
 
+  let options = $state({
+    key: 'all',
+    children: [
+      {
+        key: 'dyr',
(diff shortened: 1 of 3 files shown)


mt-kit/core/svelte/src/stories/form/CheckboxWithSubSets.stories.svelte
```diff
@@ -5,81 +5,83 @@
...
+        displayName: 'Kosmetikk',
+        docCount: 1,
+        children: []
+      }
+    ]
+  })
+
+  const { Story } = defineMeta({
+    title: 'Components/Form/CheckboxWithSubsets',
+    args: {
```
petterkoding, line 80: "you can add disableJs: false", to make it possible to toggle JS in the story

petterkoding (approved):
Looks good, just left a comment about adding disableJs so that you can toggle the JS in the story panel

image

mt-kit/core/svelte/src/stories/form/CheckboxWithSubSets.stories.svelte
kajsaeggum: Ah ofc, thx!

lha047 (approved):
👍

PR #1210, opened 2025-01-13
MTP-2803 config to include component prop types in package

kajsaeggum:
Apparently svelte > 5 is needed as a peerDependency, for component prop type info to be included in the published package. Ref https://github.com/sveltejs/kit/issues/12972

With this, the web-project "sees" what type of props the designsystem components have and should have, eliminating the lint errors of type "xxx is not assignable to type never"


lha047 (approved):
Very nice 👍

PR #1224, opened 2025-02-11
fix(MTP-2782): use text color granskog and font mattilsynet sans

eirikbacker:
- ✅ Tiny cleanup of empty `_fonts.scss` file
- ✅ Letting `@mattilsynet/design` set font-family, background and color
- ✅ Keeping local `color-mt-text-dark` for now as it is used in several components
- 🤔 Question: There is a `data-theme="green"`, but from what I can see, it is only used once; is this something we should just move into `mt-body` maybe?

### DIFF
--- a/mt-kit/core/css/src/global/_body.scss
+++ b/mt-kit/core/css/src/global/_body.scss
@@ -460,9 +460,6 @@
   display: grid;
   grid-template-rows: auto 1fr auto;
   row-gap: 0;
-  background-color: var(--body-background);
-  color: var(--body-color);
-  font-family: $font-family;
   font-size: $font-size-body;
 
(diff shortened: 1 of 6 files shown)


mt-kit/core/css/src/global/_body.scss
```diff
@@ -460,9 +460,6 @@
   display: grid;
   grid-template-rows: auto 1fr auto;
   row-gap: 0;
-  background-color: var(--body-background);
-  color: var(--body-color);
-  font-family: $font-family;
```
lha047, line 465: Great 👍 Can you also clean up the `$font-family`in _theme.scss?

lha047 (approved):
Nice 👍

PR #1226, opened 2025-02-14
fix(MTP-2782): breadcrumbs using mattilsynet/design

eirikbacker:
- Added some focus handling for enhanced accessibility
- Using breadcrumbs in `data-size="sm"` to better match original design
- Simplified logic a tiny bit
- ✅  Tested in `mattilsynet-web`


petterkoding (approved):
LGTM 👍🏼

PR #1232, opened 2025-02-18
MTP-2978 radio input

petterkoding:
- Removed `theme="button"` as this was not in use, and can easily be created by `Chip` from `@mattilsynet/design` if needed later on


lha047 (commented):
Good work 🌟

lha047 (approved):

PR #1237, opened 2025-02-24
MTP-2958 delete Fieldset component and use styles.fieldset on html tags

petterkoding:
Remove Fieldset component since its almost not in use and just use html fieldset with styles.fieldset styling instead


eirikbacker:
Duplicate of https://github.com/Mattilsynet/mattilsynet-web/pull/1332

